Issue #29, opened 2018-08-15
create DevOps Guideline


falkvoigt:
We as OIH partners need some DevOps guideline to have a same understanding of how service are setup, build and deployed

**Acceptance Criteria:**
- [ ] reviewed dokument as a Guideline for all Partners

PR #30, opened 2018-08-15
#29 add DevOps Guideline to Monorepo

falkvoigt:
Please have a look into the Document and feel free to add some Comments if something missing


philecs:
Anyone specific who should review this PR?

falkvoigt:
I would say no one specific. It's a doc perhabs just a short read should be fine?

philecs:
I think some more specifications have to be made e.g. for CI Tools and build pipelines (see #72 )

hschmidthh:
Thumbs up. Franz wrote a styleguide in the architecture repo and there are some further points for the guideline. We should merge both, or not?

philecs:
Correct - Franz wrote another styleguide. In the beginning we talked about keeping it seperate but I would suggest to create a folder with all guides in it.

philecs:
I also moved the styleguide to the monorepo (see changed files within this pull request).
@sachmerz I added you to also check the styleguide

sachmerz (approved):

hschmidthh (approved):

Guidelines/serviceOperations.md
```diff
@@ -0,0 +1,85 @@
...
+
+For OHI we decided to use small services or microservices (there is a bunch of interpretation here).
+
+Some rules to take care of:
+
+* The service should be stateless so it could be moved any time from the cluster scheduler
+* if there is an error thrown which could not be catched somewhere the service needs to be restarted by K8S
+* keep an eye on the resources. Services with high RAM or CPU usage are not easy to spread in a cluster ENV.
+* The service should be able to run multiple times on a single node (if for a good reason that is not possible the cluster should care of the affinity)
+
```
anton-kotenko, line 85: I think we should also define one more section: Reliability. 
General idea: In production environment every deployment should run in more then one instance.
This makes:
1. the whole platform won't fail in case if one deployment instances dies (e. g. just error in service, migration from one k8s node to another)
2. This make easier blue-green deployment strategy -- so no downtime during release. 

In case if it's hard to make service to run in parallel, there should be main instance and it's hot standby (e. g. use zookeeper/etcd/whatever to solve who is main and who is standby)
This requirement also may be relaxed if platform don't crash if one of services fails (e. g. some batch processing service. Dead service only introduce some additional delay)

Clearly that design solution from Scaling section often  solve problems described in Reliability section

Guidelines/styleGuide.md
```diff
@@ -0,0 +1,47 @@
...
+Use tdd (https://en.wikipedia.org/wiki/Test-driven_development)
+
+### clean code / clean architecture (recommendation)
+* https://gist.github.com/ygrenzinger/14812a56b9221c9feca0b3621518635b#file-cleanarchitecture-md
+* https://8thlight.com/blog/uncle-bob/2012/08/13/the-clean-architecture.html
+
+
+## javascript
+### node.js docker
+Use node.js docker images based on LTS version (currently node:carbon-alpine, node:carbon) instead of custom from scratch built containers.
```
anton-kotenko, line 20: Should we enforce usage of yarn instead of npm?
As far as I see in code, yarn is main tool for dependency management in monorepo.
And as far as I understand this removes package-lock.json files from services and there would be only one yarn.lock for whole project.
But this makes docker build more complex  as yarn.lock is above service folder in folders hierarchy.

anton-kotenko (approved):

Guidelines/serviceOperations.md
falkvoigt: Sure you are correct. I don't mention that in the Document because i was the opinion that is default :) i will add that thx

Guidelines/styleGuide.md
fdegenhr: We planned to use lerna for monorepo workspace support. If this is feasable, I dont see any advantage in yarn over npm.
If we choose lerna, I vote for using npm instead of yarn.

falkvoigt: We use yarn workspace in combination with lerna. Which is common practice in mono Repos.

falkvoigt:
Whats missing to get this Merged?

PR #87, opened 2018-09-03
Contributing guidelines

philecs:


jhorbulyk:
I see we have [contribution guidelines for bugs and issues](https://github.com/openintegrationhub/openintegrationhub/blob/f3575dc94043576846061f460af5d9b416d6851d/CONTRIBUTING.md#issues).  What is the contribution guidelines for raising a discussion about an open/unsolved problem?  Is it preferable to have these discussions in GitHub issues or is there an alternate preferred means?

philecs:
Good point. Are you referring to a specifc issue? @jhorbulyk 
From my pov it is preferable to discuss it in issues.

jhorbulyk:
I don't have an issue at the moment.  I think issues is a good location for discussion but it feels like a third type of item is required in the list currently containing `Bug Reports` and `Open Tasks`.

Issue #93, opened 2018-09-18
Cleanup icr service


philecs:
As a quality manager I want a properly structured repository, so that I can start the build process.

**Further description:**
In order to pass the build process we need to restructure the icr service folder:

- [x] Delete mongo folder from services/integration-content-repository folder
- [x] Delete nginx-reverse-proxy folder from services/integration-content-repository folder
- [x] Delete nginx-server folder from services/integration-content-repository folder
- [x] Delete ubuntu-base folder from services/integration-content-repository folder
- [x] Move all files from icr-api folder one level up
- [x] Add yarn

---

**Acceptance Criteria:**
- [x] All todos fulfilled

falkvoigt:
Build on Travis would work if useing the correct Docker image:
Step 1/11 : FROM icr_ubuntubase
pull access denied for icr_ubuntubase, repository does not exist or may require 'docker login'
child_process.js:645
    throw err;
    ^

falkvoigt:
As defined in Call @hschmidthh will update dockerfile and commit other changes later on

Issue #32, opened 2018-08-15
Operator should detect changes on Flow CDRs


drobiazko:
Imagine you created a new version of your flow definition which you want to deploy via `kubctl`, REST API or a frontent. It may happen that the `flow operator` will not notice the update at all. This can be easily reproduced by executing the following command for an existing flow.

`kubctl delete flow $FLOW_ID  && kubectl create flow -f $FLOW_JSON_FILE`

The changes will not be applied to the flow as the flow operator is not detecting the changes on existing `Flow` CRDs. Perfectly the flow operator checks the `metadata.resourceVersion` of the CDRs to detect changes.

Please check the paragraph **Efficient detection of changes** in the [k8s API Concepts](https://kubernetes.io/docs/reference/using-api/api-concepts/) document. Here is some extract:

> To enable clients to build a model of the current state of a cluster, all Kubernetes object resource types are required to support consistent lists and an incremental change notification feed called a watch. Every Kubernetes object has a resourceVersion field representing the version of that resource as stored in the underlying database. When retrieving a collection of resources (either namespace or cluster scoped), the response from the server will contain a resourceVersion value that can be used to initiate a watch against the server. The server will return all changes (creates, deletes, and updates) that occur after the supplied resourceVersion. This allows a client to fetch the current state and then watch for changes without missing any updates. If the client watch is disconnected they can restart a new watch from the last returned resourceVersion, or perform a new collection request and begin again.

anton-kotenko:
Requires additional investigation:
Idea: there should be flow definition crd. There also additional information should be stored for internals of platform (rabbimq accounts for flows, current state, started/stopped/running)
This additional information should be used to find out changes. To understand that flow  should be killed and all related things are deleted.
Options: 
1. add some read-only properties in flow definition crd. Need to find out how to protect them from external user
2. add "shadow" crd that is used only by platform internally. 

Need also understand how to handle crd removal for both options and trigger cleaning on it.

philecs:
Statement copied from discussion in #33 :

#32 is in progress, but It's quite complex task so It will take some time to solve the issue.

philecs:
@anton-kotenko as stated with merge of pr #104 this issue as well as  #24 #26 and #28 can be moved to "done", right? If yes, I'll put them in the right lane.

anton-kotenko:
@philecs 
You are right. Thank you.

Issue #28, opened 2018-08-15
There may be a race condition between scheduler and flow operator


drobiazko:
When the scheduler schedules a trigger execution, it does not know if the job is already ready to run (e.g. if the queues are created by the operator). Due to this race condition it may happen that the 1st scheduler tick is lost

philecs:
Statement copied from discussion in #33 :

#32 is in progress, but It's quite complex task so It will take some time to solve the issue.
#28 is convenient to solve along with #32 so it's also in progress

Issue #26, opened 2018-08-15
Remove copy & paste of CRD definition and creation


drobiazko:
There are multiple services with a copy & paste for Flow CRD definition and creation in k8s.

# Flow operator

[CRD definition](https://github.com/openintegrationhub/openintegrationhub/blob/master/services/flows-operator/index.js#L15-L33) and [CRD creation](https://github.com/openintegrationhub/openintegrationhub/blob/master/services/flows-operator/index.js#L283).

# Scheduler

[CRD definition](https://github.com/openintegrationhub/openintegrationhub/blob/master/services/scheduler/index.js#L14-L32) and [CRD creation](https://github.com/openintegrationhub/openintegrationhub/blob/master/services/scheduler/index.js#L175).

# Communication Router

[CRD definition](https://github.com/openintegrationhub/openintegrationhub/blob/master/services/communication-router/index.js#L15-L33) and [CRD creation](https://github.com/openintegrationhub/openintegrationhub/blob/master/services/communication-router/index.js#L192).

philecs:
Statement copied from discussion in #33 :

#26 is implemented in "draft mode". So I've unblocked myself for work on #32

PR #104, opened 2018-09-21
Openintegrationhub 32

anton-kotenko:
PR that solves issued #32  #24 #26 #28


scripts/build.js
```diff
@@ -1,19 +1,25 @@
 const { execSync } = require('child_process');
-const services = require(`${__dirname}/toBuild.json`);
+//FIXME have no access to integration-content-repository. Temporary exclude it from build
```
anton-kotenko: Fixme revert back when integration-content-repository build will be fixed. Details: #93

falkvoigt: That will remove all Services from the build

falkvoigt: If you want to remove icr from build make a condition in the check for changes script!

Or remove it from the Object

anton-kotenko: Sorry, but I don't understand this:
> That will remove all Services from the build

This actually removes only icr service (not all services) from build, as it is broken and can not be built until #93 will be fixed. (example of build logs in travis https://travis-ci.org/openintegrationhub/openintegrationhub/jobs/432142791)

But if you think that it would be better to address this issue in scripts/getChanges.js -- I don't mind and make make it in this way.

Generally I'd prefer to wait until #93 will be fixed before merge this PR, so it won't be required to merge some strange changes and then in few days revert them back, but this depends on plans to fix #93. @falkvoigt  do you know smth about it?

scripts/build.js
```diff
@@ -1,19 +1,25 @@
...
 
         execSync(`cd ${__dirname}/../services/${service.name}/ && yarn build`, { stdio: [0, 1, 2] });
-        execSync('echo $DOCKER_PASSWORD | docker login -u $DOCKER_USERNAME --password-stdin', { stdio: [0, 1, 2] });
-        execSync(`cd ${__dirname}/../services/${service.name}/ && docker build -t openintegrationhub/${service.name}:${service.version} .`, { stdio: [0, 1, 2] });
+        execSync(`cd ${__dirname}/../services/${service.name}/ && yarn run build:docker`, { stdio: [0, 1, 2], env: Object.assign({VERSION: service.version}, process.env) });
         execSync(`docker tag openintegrationhub/${service.name}:${service.version} openintegrationhub/${service.name}:latest`, { stdio: [0, 1, 2] });
 
-        execSync(`docker push openintegrationhub/${service.name}:${service.version}`, { stdio: [0, 1, 2] });
-        execSync(`docker push openintegrationhub/${service.name}:latest`, { stdio: [0, 1, 2] });
+        //execSync(`docker push openintegrationhub/${service.name}:${service.version}`, { stdio: [0, 1, 2] });
```
anton-kotenko: If you are speaking about this changes, it also should be reverted. But I'd like to discuss build process in general: I don't think that it's good idea to set `latest` tag to any successfully built PR's docker image, because it increases risks that somebody deploys unstable version somewhere.

scripts/build.js
falkvoigt: #93 is fixed already, but the dockerfile needs to be reworkt in icr.

Latest is only set on master builds. Because build and deploy only hits master branches.

I will review the changes tomorrow. Perhabs i missed something on my mobil ;) will come back on it

scripts/build.js
falkvoigt: That latest is only on master branch. 
The deployment will not use latest. The latest tag is just a marker for the latest image not of stable or unstable.
But as i‘m not using it feel free to remove the latest tag ;)

And i tested the Pipeline on iam without Problems.

As i said in the other post i need a big screen to Review that changes. Could be i missed something.

falkvoigt: Hi,
so i had a look on it. I did not see the hole change you did yesterday. 
Thats fine ! But you disable the Push to DockerHub completely. So we don't get any image to deploy to Cluster. (I just get why you did that) We need to think about a solution here. same for Point 1.

That Latest tag i could understand we could move that tagging to the sage comming up after Deploy in the future (integration Testing not yet implemented).

But: 
1. Why are you disabling my gcloud installation? (just see why you disable it you are not useing gcloud right? but we can't merge that in OIH because we are relating on that installation for deployment)
2. Could you please write the K8S stuff in yaml? that is more common in the Comunity.
3. you added src/ as workspace in yarn could you please name it so every partner could use /src as a lib dir ? (src/backendCommonLib or something)

scripts/build.js
drobiazko: @falkvoigt #93 is open yet and it's blocking this pr. What can we do to proceed?

falkvoigt: we need to merge the changes @hschmidthh did in his Branch an it's fine. @drobiazko please have a look at my comment below. there are some changes you did in the repo which needs some attention.

scripts/build.js
anton-kotenko: Hi, @falkvoigt.
Let me explain the reason why i did all the changes related to build process: The reason is one and simple: build process starts on every branch (including my pull request, not only master). But build process has next problems, that makes impossible for my PR to pass tests on travis
1. #93 private base image for  integration-content-repository
2. result of every build is pushed into dockerhub with latest tag (even in case of pull request). It's a) dangerous as it increases changes to deploy some incorrect b) there is no access for me to publish 
3. There is deploy phase in all travis scenario. It runs script, that just install gcloud and has commented kubectl commands. But it does not work for me, as it requires some gcloud keys in environment.

So at the moment, there is a problem, that build works only for limited set of users and this actually contradicts open-source work strategy.
Due to reasons written above I had to disable some functionality inside build scripts.

Generally I see two ways to solve all this:
1. I wait until all this issues will be fixed and then merge my PR. Then all this changes won't be required and i return back
2. Merge this PR with all this changes and then fix this issues. 
Your opinion?  

About yaml/json: I prefer json's for next reasons: we are working with javascript, so it's much easier to work with json then with yaml. If you insist on using yaml -- I can transform json descriptors into yaml, but I prefer to do this as separate task.

About src folder: Ok, I'll rename  src/lib to src/backendCommonLib

falkvoigt: Hi , @anton-kotenko 
the problem is relate to you fork the hole mono repo and also duplicate the travis pipeline. thats why you miss some informations. 

did you read my comment completely? 

1. yes Build breaks related to the Dockerfile fix is done in a different branch needs to be merged

2. i am fine with removing the latest tag for now until we have integration test an Production deployment. But push the build tag is needed for OIH Project.

3. That is related to your fork but OIH relies on this for deployment in test Cluster you can't get rid of it. what you can do , is make a condition relating github Projekt or simple do branching in OIH github instead of forking the hole Project.

We should go with the Community related to Kubernetes in our Github repo. So yaml is prefered Format here. Also a part of the DevOps Guideline which waiting in PR at the moment. 

Over all, the Problem is you forked the Project. We could not merge your PR aslong that commits related to google Cloud and docker push are in there. Which will disable 50% of the Build/deployment Pipeline.

services/communication-router/src/HttpApi.js
```diff
@@ -0,0 +1,110 @@
...
+        this._app = express();
+        this._app.use(bodyParser.json({
+            limit: 1024 * 1024
+        }));
+        this._app.use(bodyParser.urlencoded({
+            limit: 1024 * 1024,
+            extended: true
+        }));
+
+        this._app.get('/flow/:flowId', this._handleHook.bind(this));
```
pnedelko, line 23: I wanted to suggest to do it with `express Router`, but probably it would be too much for now.

scripts/build.js
anton-kotenko: Sorry, but at this moment:
1) I can not  push new branch into original repo because of lack of privileges, so i have to fork repo, with all consequences (like broken travis pipeline)
2) My build fails in travis without this changes.
What should i do? I suppose that any other developer who is not member of this repo will face the  same problems. (is this open-source?) 

About docker push of latest/current version: I have no privileges to push that images at all in openintegrationhub repo. So If i return any push, build will just fail.

About gcloud  installation and k8s. At the moment (in master branch) script just installs gcloud cli and does nothing. So what i've commented does not changes anything in build pipeline.  

So there is one more option: I revert back all changes in build pipeline, but we ignore broken build when merging PR (and repo settings should be adjusted for this)   

> We should go with the Community related to Kubernetes in our Github repo. So yaml is prefered Format here. Also a part of the DevOps Guideline which waiting in PR at the moment.

Ok. I understand this argument.

services/communication-router/src/HttpApi.js
```diff
@@ -0,0 +1,110 @@
...
+        this._logger.info({port: listenPort}, 'Going to listen for http');
+        this._app.listen(listenPort);
+    }
+
+    async _handleHook(req, res) {
+        this._logger.trace(req.params, 'hook message');
+        try {
+            const flow = await this._crdClient.flows(req.params.flowId).get();
+            if (!flow) {
+                throw new Error('404');
```
pnedelko: Maybe more meaningful error message? How about custom http errors?

services/communication-router/src/HttpApi.js
```diff
@@ -0,0 +1,110 @@
...
+            }
+            const flowModel = new Flow(flow.body);
+            const step = flowModel.getFirstNode();
+            if (!step) {
+                throw new Error('404');
```
pnedelko: More meaningful error message would be nice.

services/communication-router/src/HttpApi.js
```diff
@@ -0,0 +1,110 @@
...
+                queue,
+                Buffer.from(JSON.stringify(msg)),
+                {
+                    headers:  scheduleRecord
+                }
+            );
+
+            res.status(200);
+            res.set('Content-Type', 'application/json');
+            res.end(JSON.stringify({status: 'OK'}));
```
pnedelko: I propose to use `res.json` instead of `res.set` + `res.end`.

scripts/build.js
drobiazko: 👍 for the open source argument @anton-kotenko. ATM it looks like @falkvoigt is the only one enjoying the privilege to build the project successfully. This sucks big time and not how an open source project must work.

services/communication-router/src/HttpApi.js
```diff
@@ -0,0 +1,110 @@
...
+        } catch (e) {
+            this._logger.error(e, req.params, 'hook request');
+            res.status(500);
+            res.end();
```
pnedelko: we need to show some message to a user

scripts/build.js
falkvoigt: We need to give permission for you or we make a condition related on Project name or id.

services/communication-router/src/HttpApi.js
```diff
@@ -0,0 +1,110 @@
...
+    _getRequestData(flowId, req) {
+        const REQUEST_FIELDS = [
+            'headers',
+            'url',
+            'method',
+            'originalUrl',
+            'query',
+            'body'
+        ];
+        const msg = {
```
pnedelko, line 81: We need some helper in order to generate this kind of default message structure.

scripts/build.js
drobiazko: @falkvoigt this is an open source project. You would need to give the permissions to everybody on the globe.

falkvoigt: I can understand the fact with lag on permissions. But is it Open Source to disable Project related stuff locally and try to push it? And destroy the Pipeline which is only for OIH and not someone who wants to fork it?

it crashes on you side because you try to use Project internals. Normaly that is not the case when useing the OIH environment and not the forked one.

falkvoigt: not for every thing but nearly yes. I would restrict direct commiting to master. the other stuff i'm with you.

services/communication-router/src/HttpApi.js
```diff
@@ -0,0 +1,110 @@
...
+        }
+        if (typeof msg.body === 'string' || Buffer.isBuffer(msg.body)) {
+            // Body should be JSON object and not string and not buffer
+            msg.body = {};
+        }
+        if (req.method === 'GET') {
+            msg.body = req.query;
+        }
+
+        msg.taskId = flowId;
```
pnedelko, line 101: Is this property required?

pnedelko: Is it used somewhere afterwards?

services/flows-operator/index.js
```diff
@@ -1,313 +1,11 @@
...
 (async () => {
-    const logger = bunyan({name: 'startup', level: LOG_LEVEL});
     try {
-        logger.info('Starting platform');
-        await init(logger);
+        const app = new FlowOperatorApp();
+        await app.start();
     } catch (e) {
-        logger.error(e, 'failed to start platform');
+        console.error('Critical error, going to die', e, e && e.stack); //eslint-disable-line
```
pnedelko, line 8: why `console.error` instead of `logger`?

scripts/build.js
falkvoigt: @drobiazko i vote for open branch for every one restrict access for master. I'm not the github admin to do so. I just comment that the PR will brake pipeline on Project side.

anton-kotenko: About open-source: At  this moment external developer is not able to create PR and successfully pass tests.  So it's a problem of test+build pipeline or general work process organization. 
I don't think that my solution of the problem is the best, but problem exists. And we should do something with this. 
So we have a question: how to avoid all the internals (build, docker push, gcloud) for external users.

falkvoigt: i already answered that.
2 different approches:
1. git branch create permission to every one and don't try to PR from a fork
2. make a condition based on Projectname/id to the push and google stuff

services/flows-operator/src/FlowOperator.js
```diff
@@ -0,0 +1,293 @@
...
+        for (let job of allJobs) {
+            const flowId = this._getFlowIdFromJob(job);
+            const stepId = this._getStepIdFromJob(job);
+            if (!flowsIndex[flowId] || !flowsIndex[flowId][stepId]) {
+                await this._undeployJob(job);    
+            }        
+        }
+    }
+
+    _buildMQInedx(queues, exchanges, bindings) {
```
pnedelko, line 156: Typo: Inedx => Index

falkvoigt:
hi i just had a call with @drobiazko  related to that.
we just merge the PR and i will care on the Contitions related to the Pipeline

services/flows-operator/src/FlowOperator.js
```diff
@@ -0,0 +1,293 @@
...
+    _buildDescriptor(flow, step, queues) {
+        let jobName = flow.id +'.'+ step.id;
+        jobName = jobName.toLowerCase().replace(/[^0-9a-z]/g, '');
+        const env = this._prepareEnvVars(flow, step, queues[step.id]);
+        return this._generateAppDefinition(flow, jobName, env, step);
+    }
+
+    async _deployStep(flow, step, queues) {
+        let jobName = flow.id +'.'+ step.id;
+        jobName = jobName.toLowerCase().replace(/[^0-9a-z]/g, '');
```
pnedelko: This like and the line above is a C&P

services/flows-operator/src/HttpApi.js
```diff
@@ -0,0 +1,50 @@
...
+                throw new Error('404');
+            }
+            const flowModel = new Flow(flow.body);
+            const step = flowModel.getRecipeNodeByStepId(req.params.stepId);
+            if (!step) {
+                throw new Error('404');
+            }
+            res.status(200);
+            res.set('Content-Type', 'application/json');
+            res.end(JSON.stringify({
```
pnedelko: `res.json` is better

services/flows-operator/src/HttpApi.js
```diff
@@ -0,0 +1,50 @@
...
+                id: req.params.stepId,
+                function: step.function,
+                config: step.data || {}
+            }));
+        } catch (e) {
+            this._logger.error(req.params, 'Step info request');
+            res.status(500);
```
pnedelko: Would be nice to respond with some message, as well as with different http codes.

services/scheduler/src/Scheduler.js
```diff
@@ -0,0 +1,136 @@
...
+        this._config = app.getConfig();
+        loop(this._loopBody.bind(this), this._logger, 5000);
+    }
+    async _scheduleOne(flow) {
+        this._logger.trace({flowId: flow.id}, 'schedule flow tick');
+         
+        const scheduleRecord = {
+            'taskId': flow.id,
+            'execId': uuid().replace(/-/g, ''),
+            'userId': "DOES NOT MATTER"
```
pnedelko: I'm not aware of the style guides, but seems like `'` should be here instead of `"`.

anton-kotenko:
@falkvoigt  I've fixed .travis.yaml to run tests but skip builds for forked repos, and revert back  scripts/*
Is this ok?

services/scheduler/src/Scheduler.js
```diff
@@ -0,0 +1,136 @@
...
+        const flows = (await this._crdClient.flows.get()).body.items;
+        const schedulerRecords = (await this._crdClient.schedulerrecords.get()).body.items;
+        const schedulerRecordsIndex = schedulerRecords.reduce((index, record) => {
+            index[record.metadata.name] = record;
+            return index; 
+        }, {});
+        const now = new Date();
+        for (let flow of flows) {
+            const flowModel = new Flow(flow);
+            const firstStep = flowModel.getFirstNode();
```
pnedelko: Naming convention: would be nice to stick with one of the following: `node` or `step`. I would prefer `node`.

falkvoigt:
Yes this is a great solution to the Problem.
Even better than my condition in the script for sure.

services/scheduler/src/Scheduler.js
```diff
@@ -0,0 +1,136 @@
...
+                    this._logger.trace({
+                        name: schedulerRecord.metadata.name,
+                        dueExecution: schedulerRecord.spec.dueExecution
+                    }, 'Going to delete scheduler record due non-polling flow');
+                    await this._crdClient.schedulerrecords(flowModel.id).delete();
+                }
+                continue;
+            }
+            const schedulerRecord = schedulerRecordsIndex[flowModel.id] || {
+                "apiVersion": "elastic.io/v1",
```
pnedelko: again codestyle: is `"` ok here?

services/scheduler/src/Scheduler.js
```diff
@@ -0,0 +1,136 @@
...
+                "spec": {
+                }
+            };
+            delete schedulerRecordsIndex[flowModel.id];
+            if (!schedulerRecord.spec.dueExecution || now >= new Date(schedulerRecord.spec.dueExecution)) { 
+                try {
+                    await this._scheduleOne(flowModel);
+
+                    const newDueExecution = new Date();
+                    newDueExecution.setMinutes(newDueExecution.getMinutes() + 3);
```
pnedelko: Would be nice to move magic number to some constant.

pnedelko: And how about `cron` flows?

src/lib/src/K8sService.js
```diff
@@ -0,0 +1,72 @@
...
+
+const FlowCRD = require('../data/FlowCRD.json');
+const SchedulerCRD= require('../data/SchedulerCRD.json');
+
+class K8sService{
+    constructor(app) {
+        this._app = app;
+    }
+
+    async start() {
```
pnedelko: Looks to me more like `init`, cause it doesn't really establish any connection or something like that.

pnedelko (changes requested):

services/scheduler/src/Scheduler.js
anton-kotenko: cron flows are not implemented atm

pnedelko (approved):

PR #112, opened 2018-09-26
Flow Operator and Node container use different RabbitMQ credentials

### LINKED ISSUE #25: Flow operator is using the same RabbitMQ credentials for itself and for flows
The flow operator requires privileges to create queues and exchanges in the virtual host. These component needs to read and push data only

pnedelko:
Fixes #25

### DIFF
--- a/services/flows-operator/src/FlowOperator.js
+++ b/services/flows-operator/src/FlowOperator.js
@@ -395,7 +395,7 @@ class FlowOperator {
     }
 
     async _createAmqpCredentials(flow) {
-        const username = uuid.v4(); //@todo: use flow.id as a username?
+        const username = flow.id;
         const password = uuid.v4();
 
         this._logger.trace('About to create RabbitMQ user');
--- a/services/flows-operator/src/FlowOperator.js
+++ b/services/flows-operator/src/FlowOperator.js
@@ -63,6 +63,7 @@ class FlowOperator {
                     }
                 });
             }
+
             if (queuesStructure[flowId]) {
                 const channel = await this._channelPromise;
                 for (let queue of queuesStructure[flowId].queues) {
@@ -71,14 +72,15 @@ class FlowOperator {
                 for (let exchange of queuesStructure[flowId].exchanges) {
                     await channel.deleteExchange(exchange);
                 }
-                await this._deleteFlowSecret(flow);
             }
+
+            await this._deleteFlowSecret(flow);
 
             flow.metadata.finalizers = (flow.metadata.finalizers || []).filter(finalizer => finalizer !== FLOW_FINALIZER_NAME);
             //FIXME make sure 409 works. So non-sequential updates should go into next iteration
             //possibly handle revision field
             await this._crdClient.flow(flow.id).put({
-                body: flow
+                body: flow.toCRD()
             });
         } else {
             if (!(flow.metadata.finalizers || []).includes(FLOW_FINALIZER_NAME)) {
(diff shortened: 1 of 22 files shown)


philecs:
@pnedelko is this pr related to #25 ?

pnedelko:
@philecs yes, you are right

src/backendCommonsLib/package.json
```diff
@@ -13,6 +13,7 @@
     "amqplib": "0.5.2",
     "bunyan": "1.8.12",
     "kubernetes-client": "6.1.0",
+    "lodash": "^4.17.11",
```
anton-kotenko: Pls do not use ^ versions. This may cause unstability in builds and deploys.

services/flows-operator/src/FlowOperator.js
```diff
@@ -313,6 +340,81 @@ class FlowOperator {
...
+    _prepareAmqpUri({ username, password }) {
+        const baseUri = new URL(this._config.get('RABBITMQ_URI_FLOWS'));
+        baseUri.username = username;
+        baseUri.password = password;
+
+        return baseUri.toString();
+    }
+
+    async _createAmqpCredentials(flow) {
+        const username = uuid.v4(); //@todo: use flow.id as a username?
```
anton-kotenko: I think it is good idea: it would be easier to debug what's happening in case of problem, as it makes easier to find out who is initiator of connection to rabbit

services/flows-operator/src/FlowOperator.js
```diff
@@ -313,6 +340,81 @@ class FlowOperator {
...
+
+        await this._coreClient.secrets(secret.id).delete();
+        this._logger.debug('Secret has been deleted');
+    }
+
+    async _getFlowSecret(flow) {
+        try {
+            const result = await this._coreClient.secrets(flow.id).get();
+            return FlowSecret.fromDescriptor(result.body);
+        } catch (e) {
```
anton-kotenko, line 390: generally it would be better to handle 404 error and all other errors differently
As 404 is actually ok in our situation. But all other errors are errors, and we can't do nothing with it except to rethrow, let upper layers decide how to handle it.

services/flows-operator/src/FlowOperator.js
```diff
@@ -111,19 +117,29 @@ class FlowOperator {
...
+                const queues =  await this._queueCreator.makeQueuesForTheFlow(flow);
+                for (let node of flow.nodes) {
                     this._logger.trace({flow: flow.metadata.name, node: node.id}, 'Going to create flow node');
-                    await this._deployNode(flowModel, node, queues);
+                    await this._deployNode(flow, node, queues);
                 }
             } else {
                 this._logger.trace({name: flow.metadata.name}, 'Nothing changed. Ensure nodes and queues exists');
+
+                if (!jobsIndex[flowId]) {
```
anton-kotenko: I'd rather add code that ensures secret above this: just  before the logic that decides should we totally redeploy the whole task due to changes.
Because in that case we ensure secret even if task should be redeployed

services/flows-operator/src/FlowOperator.js
```diff
@@ -111,19 +117,29 @@ class FlowOperator {
...
+                    const flowSecret = await this._getFlowSecret(flow);
```
anton-kotenko: One optimization may be implemented here:
we may fetch all secrets in once and build hash-table for them instead of asking this from k8s api for every flow

src/backendCommonsLib/src/FlowSecret.js
```diff
@@ -0,0 +1,49 @@
...
+    get amqpUsername() {
+        const { username } = new URL(_.get(this, 'data.AMQP_URI'));
+        return username;
+    }
+
+    toDescriptor() {
+        return {
+            apiVersion: 'v1',
+            kind: 'Secret',
+            metadata: this.metadata,
```
anton-kotenko, line 28: Could you please use ownership mechanism like in job's metadata.
This gives us possibility to interlink secrets with flows as it's used in k8s

anton-kotenko (approved):

Issue #138, opened 2018-11-07
Finalize Flow Repository Specification


philecs:
As a _user type_, I want to _task_ so that _goal_.

**Further description:**

Current status of specification : [flow repo concept](https://app.swaggerhub.com/apis/drobiazko/flows-repository/0.0.3)

- [ ] Make sure service is matching services guidelines

---

**Acceptance Criteria:**
- [ ] Specifications meet service guidelines

---

[**Definition of Done**](https://github.com/openintegrationhub/openintegrationhub/blob/crudmonitoring/CONTRIBUTING.md#definition-of-done)

philecs:
See #144

philecs:
In the current version of [flow repo concept](https://app.swaggerhub.com/apis/drobiazko/flows-repository/0.0.3) :
- There is no [error status code for `flows`](https://github.com/openintegrationhub/openintegrationhub/blob/master/Guidelines/ServiceAPIGuidelines.md#may-any-object-may-have-many-owners) specified. Only 200 is specified.
- Status code 410 is not defined for **delete** `flows/{flowId}`

drobiazko:
thank you for comments. will apply them in 0.0.5

btw the most recent version is 0.0.4: https://app.swaggerhub.com/apis/drobiazko/flows-repository/0.0.4

drobiazko:
I think we should remove the requirement for returning 410 on delete requests. If the resource has been deleted we don't really know if it ever existed.

Which resource exactly do you mean by?

> There is no error status code for flows specified. Only 200 is specified.

philecs:
Agree with the 410 suggestion - I had similar thoughts.

> Which resource exactly do you mean by?

I'm talking about this one:
https://app.swaggerhub.com/apis/drobiazko/flows-repository/0.0.4#/flows/getFlows

There is only status code ´200´ defined. According to the guidelines we need to define a 404 for collections, in case the collection is missing.

drobiazko:
404 makes only sense for GET resources that retrieve entities by concrete IDs. We need to make it more clear in the guidelines. Can you create an issue for the 2 changes please and assign it to me?

philecs:
I already created a pull request to delete 410. We can use this PR to incorporate all changes.
Will add another issue for the GET status codes.

philecs:
https://app.swaggerhub.com/apis/drobiazko/flows-repository/0.0.5 current version

philecs:
[...]

PR #149, opened 2018-11-20
Delete 410 code for delete ops

philecs:
We don't need an error code for already deleted objects. If the resource has been deleted we don't really know if it ever existed.
Discussed in #138

### DIFF
--- a/Guidelines/ServiceAPIGuidelines.md
+++ b/Guidelines/ServiceAPIGuidelines.md
@@ -160,6 +160,7 @@ DELETE requests are used to delete resources. The semantic is best described as 
 
 * DELETE requests are usually applied to single resources, not on collection resources, as this would imply deleting the entire collection
 * successful DELETE requests will usually generate 200 (if the deleted resource is returned) or 204 (if no content is returned)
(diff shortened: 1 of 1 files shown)


Guidelines/ServiceAPIGuidelines.md
```diff
@@ -160,7 +160,6 @@ DELETE requests are used to delete resources. The semantic is best described as
 
 * DELETE requests are usually applied to single resources, not on collection resources, as this would imply deleting the entire collection
 * successful DELETE requests will usually generate 200 (if the deleted resource is returned) or 204 (if no content is returned)
-* failed DELETE requests will usually generate 404 (if the resource cannot be found) or 410 (if the resource was already deleted before)
```
drobiazko, line 163: I think we should keep this:

> failed DELETE requests will usually generate 404 (if the resource cannot be found)

drobiazko (changes requested):

Guidelines/ServiceAPIGuidelines.md
philecs: correct. Added this part again

drobiazko (approved):

PR #192, opened 2018-12-17
Resource coordinator doesn't depend on particular flows storage

pnedelko:
- Resource coordinator's npm lib moved to `src`
- Introduces `Flows DAO`
- Introduced `Execution drivers` abstraction
- Introduced `Kubernetes driver`
- Added unit tests


services/resource-coordinator/config/default.json
```diff
@@ -1,7 +1,9 @@
 {
-    "SELF_API_URI": "http://api-service.platform.svc.cluster.local:1234",
     "DEFAULT_CPU_LIMIT": "100m",
     "DEFAULT_MEM_LIMIT": "256Mi",
     "DEFAULT_CPU_REQUEST": "100m",
-    "DEFAULT_MEM_REQUEST": "256Mi"
+    "DEFAULT_MEM_REQUEST": "256Mi",
+    "RABBITMQ_MANAGEMENT_URI": "http://guest:guest@localhost:15672",
```
anton-kotenko, line 8: Url's can not be default config parameters, as they depends on installation.
I think that it's good idea to create such defaults for development environment, but at any installation it should be configured in specific way for this installation.

services/resource-coordinator/package.json
```diff
@@ -1,26 +1,29 @@
...
     "start": "node index.js",
+    "start:dev": "npm start | bunyan",
     "build": "echo \"No Build defined\"",
-    "build:docker": "docker build -t openintegrationhub/flows-operator:${VERSION} -f Dockerfile ../../"
+    "build:docker": "docker build -t openintegrationhub/resource-coordinator:${VERSION} -f Dockerfile ../../"
   },
   "author": "elastic.io",
   "license": "APL-2.0",
   "dependencies": {
+    "@openintegrationhub/resource-coordinator": "*",
```
anton-kotenko, line 17: Could you please explan how it is installed: from npm registry or  locally, from folder nearby?

services/resource-coordinator/src/drivers/kubernetes/KubernetesDriver.js
```diff
@@ -0,0 +1,203 @@
+const { BaseDriver } = require('@openintegrationhub/resource-coordinator');
```
anton-kotenko, line 1: I think that kubernetes driver is common solution enough to put it into src/resource-coordinator along with base-driver.
Anyway (at least at this moment) we do not plan to have other backends except k8s.

services/resource-coordinator/src/drivers/kubernetes/KubernetesDriver.js
```diff
@@ -0,0 +1,203 @@
...
+class KubernetesDriver extends BaseDriver {
+    constructor(config, logger, k8s) {
+        super();
+        this._config = config;
+        this._logger = logger;
+        this._coreClient = k8s.getCoreClient();
+        this._batchClient = k8s.getBatchClient();
+    }
+
+    async createApp(flow, node, envVars, secretEnvVars) {
```
anton-kotenko, line 16: What do you think about to implement smth like runCommand from e.io k8s driver instead of createApp? This solution looks cleaner and more generic: giving us better abstraction.

services/resource-coordinator/src/drivers/kubernetes/KubernetesDriver.js
```diff
@@ -0,0 +1,203 @@
...
+                            resources: this._prepareResourcesDefinition(flow, node)
+                        }]
+                    }
+                }
+            }
+        };
+    }
+
+    _prepareResourcesDefinition(flow, node) {
+        const lc = 'resources.limits.cpu';
```
anton-kotenko, line 164: Honestly, i don't like such short  var names :), but it really does not matter

services/resource-coordinator/src/drivers/kubernetes/KubernetesRunningNode.js
```diff
@@ -0,0 +1,52 @@
+const _ = require('lodash');
+const { RunningNode } = require('@openintegrationhub/resource-coordinator');
+
+class KubernetesRunningNode extends RunningNode {
```
anton-kotenko, line 4: Same as with k8s driver. Maybe this is generic enough to be placed in src/resource-coordinator?

services/resource-coordinator/src/ResourceCoordinatorApp.js
```diff
@@ -0,0 +1,62 @@
...
+            this.getQueueCreator(),
+            this.getRabbitmqManagement(),
+            this.getAmqp().getConnection(),
+            flowsDao,
+            driver
+        );
+        await flowOperator.start();
+    }
+
+    getK8s() {
```
anton-kotenko: are this methods (getXXXX) required? as far as I see all dependencies are injected as constructor arguments.

services/resource-coordinator/src/dao/FlowsK8sDao.js
```diff
@@ -0,0 +1,42 @@
...
+        return this._crdClient.flow(flow.id).put({
+            body: flow.toCRD()
+        });
+    }
+
+    async findAll() {
+        return ((await this._crdClient.flows.get()).body.items || []).map(item => new Flow(item));
+    }
+
+    async ensureFinalizer(flow) {
```
anton-kotenko, line 39: Just to think about:
maybe we may implement flow model as active record?
1) this makes  interface of flows dao smaller (only CRUD methods)  and consequently more generic
2) At least in my opinion: ensureFinializer and removeFinalizer are property of Flow itself, not DAO layer.
3) it's clear how to implement it: flowsDao is actually fabric that creates flows, so it may inject reference to itself into Flow. And then flow may implement several methods like save/update/delete on itself, that proxies changes to dao

services/resource-coordinator/src/drivers/kubernetes/KubernetesDriver.js
```diff
@@ -0,0 +1,203 @@
...
+        this._logger.debug('Secret has been created');
+
+        return flowSecret;
+    }
+
+    async destroyApp(appId) {
+        //TODO wait until job really will be deleted
+        this._logger.info({name: appId}, 'going to undeploy job from k8s');
+        try {
+            await this._batchClient.jobs.delete({
```
anton-kotenko, line 81: we need to handle 404 error here. Delete for non-existing entity is not an error.
But any other error is error. So we need to retry deletion in some way. (I assume that resource coordinator works with loop, so reties are already implemented)

src/resource-coordinator/src/ResourceCoordinator.js
```diff
@@ -0,0 +1,220 @@
...
+            return;
+        }
+
+        // ensure flow infrastructure - finalizer + secret + queues
+        await this._flowsDao.ensureFinalizer(flow); //@todo: onFlowCreated?
+
+        if (await this._isRedeployRequired(flow, appsIndex)) {
+            this._logger.trace({name: flow.id}, 'Flow changed. Redeploy.');
+            await this._deleteRunningAppsForFlow(flow, appsIndex);
+            await this._deleteQueuesForFlow(flow, queuesStructure); // delete all queues? really?
```
anton-kotenko, line 68: It's easier to remove all queues and recreate them.
Otherwise we'll have next problems
1. it's quite inconvenient to do  patch
2. old ques may contain some not processed data. If flow structure is changed we have risks that we'll feed step with some outdated messages (that was planned to put into other step)
So I don't see alternative solution.

anton-kotenko (approved):

services/resource-coordinator/package.json
pnedelko: Locally

PR #252, opened 2019-01-14
Re-added latest api version for secrets-service for review

sachmerz:


services/secret-service/doc/openapi.yaml
```diff
@@ -1,14 +1,16 @@
 openapi: 3.0.0
 info:
   description: "Secrets service is used in the Open Integration Hub for managing secrets used in integration flows."
-  version: "0.0.5"
+  version: "2.2.0"
```
drobiazko: Why not 0.0.6?

services/secret-service/doc/openapi.yaml
```diff
@@ -49,19 +51,15 @@ paths:
           application/json:
             schema:
               type: "object"
-              properties:
-                data:
-                  $ref: "#/components/schemas/MutableAuthClient"
+              $ref: "#/components/schemas/MutableAuthClient"
```
drobiazko: According to [API Guidelines](https://github.com/openintegrationhub/openintegrationhub/blob/master/Guidelines/ServiceAPIGuidelines.md) we return `data` object in the root of the response.

services/secret-service/doc/openapi.yaml
```diff
@@ -49,19 +51,15 @@ paths:
...
         200:
           description: "successful operation"
           content:
             application/json:
               schema:
                 type: "object"
-                properties:
-                  data:
-                    $ref: "#/components/schemas/AuthClient"
+                $ref: "#/components/schemas/MutableAuthClient"
```
drobiazko: See above

services/secret-service/doc/openapi.yaml
```diff
@@ -87,15 +85,11 @@ paths:
...
```
drobiazko: See above

services/secret-service/doc/openapi.yaml
```diff
@@ -114,25 +108,77 @@ paths:
           application/json:
             schema:
               type: "object"
-              properties:
-                data:
-                  $ref: "#/components/schemas/MutableAuthClient"
+              $ref: "#/components/schemas/MutableAuthClient"
```
drobiazko: Same as above

services/secret-service/doc/openapi.yaml
```diff
@@ -114,25 +108,77 @@ paths:
...
         200:
           description: "successful operation"
           content:
             application/json:
               schema:
                 type: "object"
-                properties:
-                  data:
-                    $ref: "#/components/schemas/AuthClient"
+                $ref: "#/components/schemas/MutableAuthClient"
```
drobiazko: Same here

services/secret-service/doc/openapi.yaml
```diff
@@ -212,9 +258,7 @@ paths:
             application/json:
               schema:
                 type: "object"
-                properties:
-                  data:
-                    $ref: "#/components/schemas/Secret"
+                $ref: "#/components/schemas/Secret"
```
drobiazko: Same as above

services/secret-service/doc/openapi.yaml
```diff
@@ -239,19 +283,15 @@ paths:
           application/json:
             schema:
               type: "object"
-              properties:
-                data:
-                  $ref: "#/components/schemas/MutableSecret"
+              $ref: "#/components/schemas/MutableSecret"
```
drobiazko: Same

services/secret-service/doc/openapi.yaml
```diff
@@ -239,19 +283,15 @@ paths:
...
         200:
           description: "successful operation"
           content:
             application/json:
               schema:
                 type: "object"
-                properties:
-                  data:
-                    $ref: "#/components/schemas/Secret"
+                $ref: "#/components/schemas/Secret"
```
drobiazko: See above

services/secret-service/doc/openapi.yaml
```diff
@@ -299,22 +339,87 @@ components:
...
+      }
     MutableOAuth1Client:
       required:
         - consumerKey
         - consumerSecret
         - nonce
         - signature
         - signatureMethod
-        - verifier
+        - version
```
drobiazko: `verifier` stands for the `oauth_verifier` param defined in the OAuth1 spec. It was replaced by `version`. Why is and what stands `version` for?

services/secret-service/doc/openapi.yaml
```diff
@@ -331,9 +436,45 @@ components:
         version:
           type: string
     MutableAuthClient:
+      required:
+        - name
+        - owners
```
drobiazko, line 417: The property `owners` should belong to `AuthClient`, not `MutableAuthClient`

services/secret-service/doc/openapi.yaml
```diff
@@ -299,22 +339,87 @@ components:
...
+        "_id": "5c06ee51120e789235d9dc19",
+        "endpoint": {
+          "auth": "https://accounts.google.com/o/oauth2/v2/auth?prompt=consent&scope={{scope}}&access_type=offline&include_granted_scopes=true&state={{state}}&redirect_uri={{redirectUri}}&response_type=code&client_id={{clientId}}",
+          "token": "https://www.googleapis.com/oauth2/v4/token",
+          "userinfo": "https://www.googleapis.com/oauth2/v4/userinfo"
+        },
+        "clientId": "myclientid",
+        "clientSecret": "shhhhhhhh",
+        "redirectUri": "http://localhost:9001/api/v1/callback",
+        "mappings": {
```
drobiazko: Why do we need this?

services/secret-service/doc/openapi.yaml
```diff
@@ -299,22 +339,87 @@ components:
...
-        name:
-          type: string
         clientId:
           type: string
         clientSecret:
           type: string
-        refreshTokenUrl:
+        redirectUri:
           type: string
+        endpoint:
```
drobiazko: What is the purpose of this object?

services/secret-service/doc/openapi.yaml
```diff
@@ -299,22 +339,87 @@ components:
         - clientSecret
         - refreshTokenUrl
       properties:
-        name:
-          type: string
         clientId:
           type: string
         clientSecret:
           type: string
-        refreshTokenUrl:
```
drobiazko, line 308: Why was it removed? We have a requirement to refresh tokens automatically.

services/secret-service/doc/openapi.yaml
```diff
@@ -482,13 +646,14 @@ components:
     Owner:
       type: object
       required:
-        - id
-        - type
+        - entityId
```
drobiazko: This change violates [API Spec Guidelines](https://github.com/openintegrationhub/openintegrationhub/blob/master/Guidelines/ServiceAPIGuidelines.md).

drobiazko (changes requested):

services/secret-service/doc/openapi.yaml
sachmerz: I believe we have agreed in one of the meetings with CES, that our services start with version 1.0.0. As we already have this service running, for us these changes are the next iteration, thus v. 2.x

services/secret-service/doc/openapi.yaml
sachmerz: It looks superfluous for a single object and we haven't discussed any meta properties on a single object. But since we have agreed in the API guidelines, we will modify the Response objects.

services/secret-service/doc/openapi.yaml
sachmerz: You have `verifier` as required, but `version` on line 331 in your version v.0.0.5. Probably a typo.
I guess you meant verifier in both cases? Will be `verifier` in the next commit.

services/secret-service/doc/openapi.yaml
sachmerz: Why not? Why do you have `owners` then in you initial MutableSecret? And how should owners be modified, if not through e.g. a patch request?

services/secret-service/doc/openapi.yaml
sachmerz: Which line exactly? If you mean `mappings` then it is optional but we actually need it to find or fetch the user reference. We are still working on it to minimize the API footprint.
You can ignore it, if you do not need it.

services/secret-service/doc/openapi.yaml
sachmerz: We need authentication and token refreshing endpoints. These are grouped in this object.

services/secret-service/doc/openapi.yaml
sachmerz: It is moved to `endpoints.token`, see above.

services/secret-service/doc/openapi.yaml
sachmerz: We had an issue in the past when defining `id`, `type` in a schema. We'll check this one and align it to api spec.

services/secret-service/doc/openapi.yaml
drobiazko: Aren't we talking about the spec version? we don't care about the implementation version here.

services/secret-service/doc/openapi.yaml
drobiazko: You are right. My fault.

services/secret-service/doc/openapi.yaml
drobiazko: An implementation may return more properties in the response but do we really need to expose them in the spec? You would force us to return them as well event though they make sense for your specific implementation only.

services/secret-service/doc/openapi.yaml
drobiazko: The naming is not self-explanatory. The name `refreshTokenUrl` tells you exactly what the resource is for.

services/secret-service/doc/openapi.yaml
drobiazko: The names of this object don't match the names in the OAuth2 spec. What are these properties for:

* userinfo
* endSession

And what is the match in the OAuth2 spec?

drobiazko: I would propose to stick to a flat structure and give url properties speaking names `fooUrl` such as `refreshTokenUrl`, `revokeTokenUrl`, etc.

sachmerz: Section `3. Protocol endpoints` of [OAuth2 spec ](https://tools.ietf.org/html/rfc6749#page-18) mentions `Authorization endpoint` and `Token endpoint`, albeit the implementation details and naming of these endpoints is left to the protocol implementers. We can make these properties more speaking, e.g. `authUrl` and `tokenUrl`, but according to the spec, [token endpoint](https://tools.ietf.org/html/rfc6749#section-3.2) is responsible - among others - for refresh tokens. 
```
The token endpoint is used by the client to obtain an access token by
   presenting its authorization grant or refresh token.  The token
   endpoint is used with every authorization grant except for the
   implicit grant type (since an access token is issued directly).
```
Still, we think it makes sense to group authorization and token endpoints in one object, instead of a flat structure.
`userinfo` is meant to fetch user data. Let's say the user creates two Google/Microsoft/Atlassian secrets through the authClient and the secret contains only the tokens. Accessing userinfo would fetch username/email, so that the these can be displayed to a user through UI.
Similar to the `mappings` property, we could remove the additional endpoints (userinfo, revoke) from the spec and add them later, if needed. Our implementation of secrets-service will still support these features.

services/secret-service/doc/openapi.yaml
sachmerz: It's ok with us removing the mappings from the spec.

services/secret-service/doc/openapi.yaml
heggert: Most providers have the same URL for code and refresh token requests. Using token only as identifier (see endpoints) is more generic

services/secret-service/doc/openapi.yaml
heggert: I would say the changes are at least minor so what about 0.1.0 ?

drobiazko: ok

services/secret-service/doc/openapi.yaml
drobiazko: Ok, let's agree on the following structure?

```yaml
endpoints:
  type: object
  properties:
    auth:
      type: string
    token:
      type: string
  required:
    - auth
    - token
```
Changes:

* endpoints pluralised
* only official supported endpoints remained

sachmerz: Ok

philecs:
@sachmerz what are the next steps here?

PR #253, opened 2019-01-17
InfrastructureManager independent from Driver

pnedelko:


services/resource-coordinator/src/drivers/kubernetes/KubernetesDriver.js
```diff
@@ -1,7 +1,7 @@
 const { BaseDriver } = require('@openintegrationhub/resource-coordinator');
 const uuid = require('uuid/v4');
 const _ = require('lodash');
-const KubernetesRunningApp = require('./KubernetesRunningNode');
+const KubernetesRunningNode = require('./KubernetesRunningNode');
```
anton-kotenko: I think that KubernetesRunningNode may be bad name: reason in kubernetes vocabulary `node` is physical machine used to run `kubelet` who runs set of pods

pnedelko: @anton-kotenko I agree with you. I was trying to get rid of `app`, because in my opinion it's too abstract. I want to name this thing, so that it would be clear, that it represents a flow node running in Kubernetes. I'll come with new name suggestions a bit later.

pnedelko: @anton-kotenko how about `RunningFlowNode` for the base class, and `KubernetesRunningFlowNode` for the K8s specific?

services/resource-coordinator/src/InfrastructureManager.js
```diff
@@ -47,34 +76,50 @@ class OIHInfrastructureManager extends InfrastructureManager {
...
     }
 
-    _deleteAmqpCredentials(credentials) {
-        return this._rabbitmqManagement.deleteUser(credentials);
+    _deleteRabbitMqCredentials(flow, node) {
+        return this._credentialsStore.remove(flow.id, node.id);
+    }
+
+    async _deleteRabbitMqCredentialsForFlow(flow) {
+        const flowCredentials = await this._credentialsStore.getAllForFlow(flow.id);
```
anton-kotenko, line 117: Generally here we have a problem. Scenario:
1. assume that several flows was started, all of them was assigned credentials, .... 
2. resource coordinator was restarted for some reason (fail, update, node restart, ....)
3. from now _credentialsStore is empty
4. stop  some flow
5. Consequences: as _credentialsStore is empty,  secrets won't be removed.

anton-kotenko (approved):

services/resource-coordinator/src/InfrastructureManager.js
pnedelko: @anton-kotenko created an issue to solve this problem: https://github.com/openintegrationhub/openintegrationhub/issues/259.

PR #261, opened 2019-01-24
Icr permissions

SvenHoeffler:
The ICR is now capable of directly reading the user's permissions as supplied by the IAM, instead of implicitly deriving them from the user's roles.

For the purposes of compatibility, this functionality is currently optional. By default the ICR will continue to work as before, and the new behaviour can be activated by setting the environment variable USE_PERMISSIONS to true.


philecs:
@sachmerz and @falkvoigt I think you are more aware of the permissions of IAM. Any objections against merging the PR?

PR #299, opened 2019-02-12
Event Collaboration

pnedelko:


lib/event-bus/src/EventBus.js
```diff
@@ -24,23 +24,31 @@ class EventBus {
...
+        if ('#' in this._handlers) {
+            eventHandlers.push(...this._handlers['#']);
+        }
+        if ('*' in this._handlers) {
+            eventHandlers.push(...this._handlers['*']);
+        }
 
         // allow second level wildcards
         const fragments = eventName.split('.');
-
```
jkopen: if(fragments.length < 2 && '#' in this._handlers) eventHandlers.push(this._handlers['#']);

jkopen: Suggestion: add the code above and remove lines 30-32 to make "#" behave more closely to rabbitmq standard

lib/event-bus/examples/publish-subscribe.js
```diff
@@ -0,0 +1,48 @@
...
+
+    // make connection (required for publish and for subscribe)
+    // connection will be restarted after 1s if if it gets lost
+    await eventBus.connect();
+
+    // creating a new event
+    const event = new Event({
+        headers: {
+            name: 'flow.started',
+            serviceName: 'my-service'
```
anton-kotenko: Is it possible to take serviceName from Event bus?
It seem's quite natural. And situation when you are publishing events from name of other account looks a little bit dirty, and looks like source of different problems.
In case when publish is required it's better to create other instance of EventBus with other serviceName (may be even using shared transport)

lib/event-bus/src/EventBus.js
```diff
@@ -0,0 +1,89 @@
+const assert = require('assert');
+const Transport = require('./Transport');
+const Event = require('./Event');
+
+class EventBus {
```
anton-kotenko, line 5: Am i right, that there is no possibility to use EventEmitter here because of asynchronicity of event handlers?

lib/event-bus/src/EventBus.js
```diff
@@ -0,0 +1,89 @@
...
+        }, []);
+    }
+
+    disconnect() {
+        return this._transport.disconnect();
+    }
+
+    publish(event) {
+        assert(event instanceof Event, 'event param have to be of type Event');
+        event.setServiceName(this._serviceName);
```
anton-kotenko, line 82: +1.
But it's requried to fix examples.

lib/event-bus/src/EventBus.js
```diff
@@ -0,0 +1,89 @@
...
+        return this._transport.publish(event);
+    }
+
+    subscribe(eventName, callback) {
```
anton-kotenko, line 86: As far as i see, it's not possible to subscribe to messages after connect method was called.
It's not a problem generally (though it's possible improvement), but it's not clear  when you are just using library and don't know how it works. 
To avoid such problems, I think it would be great to throw exception here, if we are trying to subscribe after `connect` method was already called

lib/event-bus/src/RabbitMqTransport.js
```diff
@@ -0,0 +1,207 @@
...
+        return this._connection.close();
+    }
+
+    async _getPublishChannel() {
+        try {
+          if (!this._publishChannel) {
+              this._publishChannel = await this._connection.createChannel();
+
+              if (this._publishChannel.connection.stream._hadError) {
+                this._logger.error('Channel error');
```
anton-kotenko: It seems that this errors should be glued

lib/event-bus/src/RabbitMqTransport.js
```diff
@@ -0,0 +1,207 @@
...
+        this._logger.error(err);
+      }
+    }
+
+    async _assertExchange(channel) {
+      try {
+        if (this._exchangeAsserted) {
+            return;
+        }
+        if(!channel || typeof channel.assertExchange === 'undefined') {
```
anton-kotenko: I think better to throw error here. It's clearly incorrect usage of library. So earlier you show this -- easier to fix consequences

lib/event-bus/src/RabbitMqTransport.js
```diff
@@ -0,0 +1,207 @@
...
+    }
+
+    reset() {
+      this._connection = false;
+      this._exchangeAsserted = false;
+      this._publishChannel = false;
+      this._subscribeChannel = false;
+    }
+
+    async retry(){
```
anton-kotenko: should this method be public?

lib/event-bus/src/RabbitMqTransport.js
```diff
@@ -0,0 +1,207 @@
...
+      try {
+        this._logger.debug('Retry Connection:');
+        await this.connect();
+
+        if(this._connection !== false && this._subscriptions.length > 0){
+          const length = this._subscriptions.length;
+          let i;
+          for(i=0; i<length; i+=1){
```
anton-kotenko: non typical solution (why not `for (let item  of this._subscriptions) { await bla-bla-bla}`?) 
But really does not matter

lib/event-bus/src/RabbitMqTransport.js
```diff
@@ -0,0 +1,207 @@
...
+        return false;
+      }
+    }
+
+    async subscribe({topics, serviceName}, callback) {
+        this._subscriptions.push({topics, serviceName, callback});
+        this.handleSubscription({topics, serviceName}, callback);
+    }
+
+    async handleSubscription({topics, serviceName}, callback) {
```
anton-kotenko: It  looks like this method should be private

lib/event-bus/src/RabbitMqTransport.js
```diff
@@ -0,0 +1,207 @@
...
+        const channel = await this._getSubscribeChannel();
+        const queueName = serviceName;
+
+        if(!channel || typeof channel.assertQueue === 'undefined') {
```
anton-kotenko: may be just `assert` to throw error when method is used incorrerctly

lib/event-bus/src/RabbitMqTransport.js
```diff
@@ -0,0 +1,207 @@
...
+
+        await channel.assertQueue(queueName);
+        await this._assertExchange(channel);
+        for (let topic of topics) {
+            await channel.bindQueue(queueName, MAIN_EXCHANGE_NAME, topic);
+        }
+
+        return channel.consume(queueName, async msg => {
+            const json = JSON.parse(msg.content.toString());
+            const event = new RabbitMqEvent(json);
```
anton-kotenko, line 196: may be freeze event, to avoid problems when subscribers change it?

lib/event-bus/src/RabbitMqTransport.js
```diff
@@ -0,0 +1,207 @@
...
+        }
+
+        await channel.assertQueue(queueName);
+        await this._assertExchange(channel);
+        for (let topic of topics) {
+            await channel.bindQueue(queueName, MAIN_EXCHANGE_NAME, topic);
+        }
+
+        return channel.consume(queueName, async msg => {
+            const json = JSON.parse(msg.content.toString());
```
anton-kotenko: try/catch over JSON.parse. We have no guarantees that event bus will always contain only `good` messages.
Also there is one more case when msg === null. It's fake message, emitted when rmq channel is closed.

lib/event-bus/src/RabbitMqTransport.js
```diff
@@ -0,0 +1,207 @@
...
+            const event = new RabbitMqEvent(json);
+            event.setChannel(channel);
+            event.setMsg(msg);
+
+            try {
+                await callback(event);
+            } catch (e) {
+                console.error(e); //eslint-disable-line
```
anton-kotenko: use logger?

services/communication-router/config/default.json
```diff
@@ -1,5 +1,6 @@
 {
   "LISTEN_PORT": 1234,
+  "MONGODB_URI": "mongodb://localhost/communication-router",
```
anton-kotenko: Better to define dev configuration and  define  mongo uri there.
URI for mongodb can not have default value

lib/event-bus/src/EventBus.js
```diff
@@ -0,0 +1,89 @@
...
+        }
+
+        if (eventHandlers.length === 0) {
+            await event.ack();
+            return this._logger.trace({event}, 'No event handler found');
+        }
+
+        for (let handler of eventHandlers) {
+            try {
+                await handler(event);
```
anton-kotenko, line 60: Not clear behavior
1. What if one of handlers call event.ack and other event.nack?
2. What if no handler call event.ack/event.nack?

services/scheduler/src/event-handlers/FlowStarted.js
```diff
@@ -0,0 +1,19 @@
+const Flow = require('../models/Flow');
+
+module.exports = ({logger}) => async (event) => {
+    try {
+        const { payload } = event;
+        const { id } = payload;
+        const flow = await Flow.findById(id);
+        if (!flow) {
```
anton-kotenko, line 8: ~~maybe we should create flow?~~ 
 ~~Otherwise, assume scenairo~~
1. ~~new flow is created in flow-repo~~
2. ~~flow is started~~ 
3. ~~clearly that we don't have this flow in scheduler's models' storage~~
4. ~~Scheduler actually ignores this message, and will never send `ticks` to it's first step~~

Flow starting is event that creates proper instance of model in storage?

services/scheduler/src/event-handlers/FlowStarting.js
```diff
@@ -0,0 +1,24 @@
+const Flow = require('../models/Flow');
+
+module.exports = ({logger}) => async (event) => {
```
anton-kotenko, line 3: May be this event should be renamed to smth like `flow.edit` or `flow.change`
Reason: generally we should send this event every time flow is changed, to make possible to update flows scheduling parameters

services/scheduler/config/default.json
```diff
@@ -1,5 +1,7 @@
 {
+    "FLOWS_PER_TICK_LIMIT": 20,
     "LISTEN_PORT": 1234,
+    "MONGODB_URI": "mongodb://localhost/scheduler",
```
anton-kotenko: It seems to me that it's bad idea to put MONGODB_URI into default config. Reason:
assume scenario: at production testing installation we've forgot to add proper mongo uri parameter in configuration.
But service will see default value, and try to start. 
In better case it just die (as it's unable to connect to mongo) and will write some good error about this.
In worst case it will retry connecting to mongo until end of times, doing nothing and logging nothing.

In case if there would not be default value for mongodb uri, config module of service will just explode with proper error: `there is no/incorrect value for mongo uri`.

services/resource-coordinator/src/dao/FlowsDao.js
```diff
@@ -0,0 +1,14 @@
+const { FlowsDao } = require('@openintegrationhub/resource-coordinator');
+const Flow = require('../models/Flow');
+
+class OIHFlowsDao extends FlowsDao {
```
anton-kotenko, line 4: As far as i understand, current concept declares that we should use separate independent flow data storage for all services, and synchronize their states using event bus. So may be it's not required to inherit from base class?
And base class should be removed at all?

services/resource-coordinator/src/dao/FlowsDao.js
```diff
@@ -0,0 +1,14 @@
...
+    /**
+     * Find all flows.
+     * @returns {Promise<Flow[]>}
+     */
+    findAll() {
```
anton-kotenko, line 9: Why not static method?

services/resource-coordinator/src/dao/FlowsDao.js
```diff
@@ -0,0 +1,14 @@
+const { FlowsDao } = require('@openintegrationhub/resource-coordinator');
```
anton-kotenko, line 1: Do we need dao layer here? 
AFAIK now we are using dedicated model for service, so why we need dao layer?

lib/event-bus/examples/publish-subscribe.js
pnedelko: It has been already done. Changed the examples.

lib/event-bus/src/EventBus.js
pnedelko: I was also thinking of improving this, so that you could add new subscriptions after connection has been done. But let's put `throw` for now.

pnedelko: Done

lib/event-bus/src/RabbitMqTransport.js
pnedelko: Done

lib/event-bus/src/RabbitMqTransport.js
pnedelko: replaced with `assert`

lib/event-bus/src/RabbitMqTransport.js
pnedelko: added `_` at the beginning

lib/event-bus/src/RabbitMqTransport.js
pnedelko: rewritten

lib/event-bus/src/RabbitMqTransport.js
pnedelko: done

lib/event-bus/src/RabbitMqTransport.js
pnedelko: done

lib/event-bus/src/RabbitMqTransport.js
pnedelko: Good catch. Added `Object.freeze` in `EventBus`, so that you need to do this in every transport.

lib/event-bus/src/RabbitMqTransport.js
pnedelko: Done

lib/event-bus/src/EventBus.js
pnedelko: fixed examples

lib/event-bus/src/RabbitMqTransport.js
pnedelko: done

services/scheduler/config/default.json
pnedelko: removed from default config

services/communication-router/config/default.json
pnedelko: removed from default config

services/scheduler/src/event-handlers/FlowStarted.js
pnedelko: Yes. On `flow.starting` we just save data to the local DB. Then when `Resource Coordinator` creates queues and deploys containers and sends `flow.started`. Scheduler receives this event and starts scheduling.

services/scheduler/src/event-handlers/FlowStarting.js
pnedelko: We decided to block any edits on running flows, as it is implemented in e.io. Therefore you will need to stop your flow first, make changes and then start again.

services/resource-coordinator/src/dao/FlowsDao.js
pnedelko: In order to have some abstraction over database?

Also related question: should the `Event Bus` be a part of the `Framework`? Currently `Resource Coordinator` does only publishing and doesn't subscribe to events.

services/resource-coordinator/src/dao/FlowsDao.js
pnedelko: Then this class would have only static methods?

services/resource-coordinator/src/dao/FlowsDao.js
pnedelko: Don't know. Need to discuss with you.

lib/event-bus/src/RabbitMqTransport.js
pnedelko: TODO: add ack (so that this message won't come again) && log headers

lib/event-bus/src/EventBus.js
pnedelko: @anton-kotenko created https://github.com/openintegrationhub/openintegrationhub/issues/324 in order to discuss possible solutions to this problem.

lib/event-bus/src/EventBus.js
pnedelko: you are right

anton-kotenko (approved):

PR #307, opened 2019-02-15
Create .eslintrc.yaml

DennisCES:
Just for better readability


philecs:
@DennisCES status here?

Issue #272, opened 2019-02-04
Expand ICR for service collaboration


philecs:
As a product manager I want to secure interoperability of framework services so that stakeholder can easily operate the oih.

**Further description:**

To ensure interoperability of the ICR API must be expanded to be able to start/stop a flow.
Therefore the two following endpoints are needed:

We need to extend the API by the following resources:

* `POST /flows/{id}/start`: Used to start a flow
* `POST /flows/{id}/stop`: Used to stop a flow

Starting a flow means setting flow's `status` to `active` and raising `flow.starting` event. Stopping a flow means setting flow's `status` to `inactive` and raising `flow.stopping` event. The schema of the event payload is shown below.

```yaml
Event:
  type: object
  required:
    - headers
    - payload
  properties:
    headers:
      type: object
       required:
         - name
         - createdAt
      properties:
        name:
          type: string
        createdAt:
          type: string
          format: date-time
    payload:
      type: object
```

The `payload` property is an arbitrary object to be sent with the event. Flow repository will send the entire flow as `payload`. 

**We also need to extend the `flow` model in the API by the following properties:**

* ` id`: required
* `cron`: required for polling flows and defaults to `*/3 * * * *` if not set for a polling flow. Must not be set for a webhook flow.

---

**Acceptance Criteria:**
- [ ] Endpoints added
- [ ] Flow model adjusted
- [ ] Test delivered
- [ ] Endpoints checked by @DennisCES 

---

[**Definition of Done**](https://github.com/openintegrationhub/openintegrationhub/blob/crudmonitoring/CONTRIBUTING.md#definition-of-done)

philecs:
Additionally see comment by @drobiazko in #262

philecs:
@SvenHoeffler: @drobiazko will update #262. The flow structure is likely to be expanded by a new attribute `cron`. If set its a scheduled flow, if not set its a webhook.

But more to come.

philecs:
#262 updated: 

We need to extend the API by the following resources:

* `POST /flows/{id}/start`: Used to start a flow
* `POST /flows/{id}/stop`: Used to stop a flow

Starting a flow means setting flow's `status` to `active` and raising `flow.starting` event. Stopping a flow means setting flow's `status` to `inactive` and raising `flow.stopping` event. The schema of the event payload is shown below.

```yaml
Event:
  type: object
  required:
    - headers
    - payload
  properties:
    headers:
      type: object
       required:
         - name
         - createdAt
[...]

Issue #262, opened 2019-01-25
Integration Services Collaboration


drobiazko:
At the moment we have following services ready and functional:

* `Scheduler`
* `Webhooks`
* `Flow Repository`
* `Resource Coordinator`

However, these services are not yet integrated. The scope of this epic is to provide integration between these services. 

The goal is to let services collaborate with each other without sending requests when some data is needed. Instead the services raise events when things change. The other services listen to these events and react appropriately. This style of collaboration is called [Event Collaboration](https://martinfowler.com/eaaDev/EventCollaboration.html). Now let's have a look at the following diagram showing the collaboration between the services mentioned above.

![oih_integration_events 1](https://user-images.githubusercontent.com/464220/51745965-df06c900-20a4-11e9-9093-1a94dd8e448b.png)

Now let's discuss the diagram above in a bit more details:

1. Client starts a flow using flow repository's REST API
2. `Flow Repository` sets the flow's `status` to `starting` and raises the event `flow.starting` 
3. There are 3 services listening to the event `flow.starting`:  Webhooks, Scheduler and Resource Coordinator. `Webhooks` and `Scheduler` examine the event's payload and decide if they need to react appropriately. We will discuss the exact reaction of both services later in this document.
4. Upon receiving `flow.starting` event the `Resource Coordinator` starts deploying the containers. Once all containers were deployed, `Resource Coordinator` raises the `flow.started` event.
5. `Flow Repository` receives the `flow.started` event and switches flow's `status` property from `starting` to `started`.
6. `Webhooks` receives the `flow.started` event and starts receiving incoming HTTP calls for the given flow.
7. `Scheduler` receives the `flow.started` event and starts scheduling the flow, according to it's cron property.
8. When a client stops a running flow using flow repository's REST API, the event `flow.stopping` is raised which is causing an inverse reaction chain of events.

Now let's discuss the individual services in details.

## Flow repository

We need to extend the API by the following resources:

* `POST /flows/{id}/start`: Used to start a flow
* `POST /flows/{id}/stop`: Used to stop a flow

Starting a flow means setting flow's `status` to `active` and raising `flow.starting` event. Stopping a flow means setting flow's `status` to `inactive` and raising `flow.stopping` event. The schema of the event payload is shown below.

```yaml
Event:
  type: object
  required:
    - headers
    - payload
  properties:
    headers:
      type: object
       required:
         - name
         - createdAt
      properties:
        name:
          type: string
        createdAt:
          type: string
[...]

PR #315, opened 2019-02-26
Flow repo collaboration

pnedelko:


services/flow-repository/app/api/controllers/startstop.js
```diff
@@ -0,0 +1,94 @@
...
+
+
+// Start a flow
+router.post('/:id/start', jsonParser, async (req, res) => {
+  const flowId = req.params.id;
+  const credentials = res.locals.credentials[1];
+  let flow;
+
+  if (!mongoose.Types.ObjectId.isValid(flowId)) {
+    return res.status(400).send('Invalid id');
```
pnedelko, line 30: The API respond sometimes with content type "application/json", sometimes with "text/plain", which is weird. But this is a general problem, not only this PR.

services/flow-repository/app/api/controllers/mongo.js
```diff
@@ -119,6 +119,60 @@ const updateFlow = (storeFlow, credentials) => new Promise((resolve) => {
...
+    qry = { '_id': findId };
+  } else {
+    qry = {
+      $and: [{ '_id': findId },
+        { 'owners.id': { $in: credentials } },
+      ],
+    };
+  }
+
+  Flow.findOneAndUpdate(
```
pnedelko, line 163: All `mongoose` methods return promises. No need to use `new Promise`.

services/flow-repository/app/api/controllers/startstop.js
```diff
@@ -0,0 +1,94 @@
...
+
+  const ev = {
+    headers: {
+      name: 'flow.starting',
+    },
+    payload: flow,
+  };
+
+  await publisher.publish(ev);
+  return res.status(200).send(ev);
```
pnedelko: Event is something internal to the system. What's the point of exposing it?

services/flow-repository/app/api/controllers/startstop.js
```diff
@@ -0,0 +1,94 @@
...
+
+  const ev = {
+    headers: {
+      name: 'flow.stopping',
+    },
+    payload: flow,
+  };
+
+  await publisher.publish(ev);
+  return res.status(200).send(ev);
```
pnedelko: Event is something internal to the system. What's the point of exposing it?

services/flow-repository/app/utils/publish.js
```diff
@@ -0,0 +1,23 @@
+const config = require('../config/index');
+const log = require('../config/logger');
+const { EventBus, RabbitMqTransport, Event } = require('../../../../lib/event-bus');
+
+async function publish(ev) {
```
pnedelko: My vision was, that an application should connect to the Event Bus on start and keep this connection till the termination. Therefore I find this code weird.

services/flow-repository/app/api/controllers/startstop.js
pnedelko: I would rather return a whole flow itself or just `id` and `status`.

pnedelko (commented):
Also the app is not subscribing to `flow.started` & `flow.stopped`.

services/flow-repository/app/utils/publish.js
SvenHoeffler: You're entirely correct. Adjusted to build a queue during startup and maintain it.

services/flow-repository/app/api/controllers/mongo.js
SvenHoeffler: Certain methods return thenables rather than full Promises, which has caused issues for us in the past. Hence why we've switched to manual wrapping.

services/flow-repository/app/api/controllers/startstop.js
SvenHoeffler: We used it internally to ensure full functionality, but you're correct that it's not necessary for actual use. Adjusted to return just id and status.

services/flow-repository/app/api/controllers/startstop.js
SvenHoeffler: We used it internally to ensure full functionality, but you're correct that it's not necessary for actual use. Adjusted to return just id and status.

services/flow-repository/app/api/controllers/startstop.js
SvenHoeffler: That's true. Do we have a shared best practice on how to format the content of such error messages?

SvenHoeffler:
> Also the app is not subscribing to `flow.started` & `flow.stopped`.

It's our understanding that the Flow Repository is only ever expected to produce events, not to consume them. Which is why, at least at this point in time, it didn't seem necessary to subscribe.

philecs:
@SvenHoeffler according to the service collaboration concept the flow repository should change the status of the flow if `flow.started` or `flow.stopped` was received.

- [Textual Description](https://github.com/openintegrationhub/openintegrationhub/issues/262#issue-403126633)
- [Graphical Representation](https://github.com/openintegrationhub/openintegrationhub/issues/262#issuecomment-465514655)

I'm not a 100% sure if the textual description already contains this information but the graphic does. 

Anything to add @pnedelko ?

SvenHoeffler:
@philecs 
According to #272 the Flow Repository receives that command through a direct POST request to its API, not an event. Changing the status of the flow is done as a response to that request. The actual event is then raised and *published* by the Flow Repo, and is then later consumed by the other services. 

Edit: Upon further review, it appears that the issue #272 refers to an outdated concept?

Edit 2: That does indeed appear to be the case. Will adjust to fit the newest graphic.

philecs:
@SvenHoeffler exactly. The information that flow repository receives a POST request as well as an event is currently not mentioned in #272 .

1. A POST request ist made so that flow repo changes status to **starting** and emits an `flow.starting` event
2. After component orchestrator started the flow it emits an event `flow.started` which is also consumed by flow repo. Flow repo itself changes the status of the flow to **active** upon receiving the event.

SvenHoeffler:
The Flow Repository will now subscribe to `flow.started` and `flow.stopped` and set the respective flow's status accordingly. 

@pnedelko One potential issue we ran into during testing: Currently, the Flow Repository will return a `nack` if changing the flow's status for whatever reason. However, in case of a missing database entry this could potentially turn into an endless loop of never-ending retries. Should clients handle this by themselves, or should the Event Bus impose an upper limit of retries?

services/flow-repository/app/api/controllers/startstop.js
pnedelko: I'm not sure, that we have such best practices defined. @philecs do we?

services/flow-repository/app/api/controllers/mongo.js
pnedelko: I guess it was quite a long time ago. Actually [this line](https://github.com/openintegrationhub/openintegrationhub/blob/master/services/integration-content-repository/app/server.js#L134) solves the problem. We are using only `await Model.method` and don't face any problems. But I'm not insisting on changing this file. Just want to share our experience with the subject.

pnedelko:
@SvenHoeffler 
> Currently, the Flow Repository will return a `nack` if changing the flow's status for whatever reason.

Do you mean you `nack` the message if you can't change the status? What are the potential reasons?

SvenHoeffler:
@pnedelko 
> Do you mean you `nack` the message if you can't change the status? What are the potential reasons?

Yes, it's based on the result of the mongoose `FindOneAndUpdate` action. The most likely causes for a `nack` are either the DB being unavailable, or that no flow with the supplied ID can be found in the database.

pnedelko:
@SvenHoeffler when DB is not available, from our point of view, it's correct to do `nack`, so that you could process this message later. As for the second case, not sure. How this situation could happen? Somebody posted a message with wrong ID? In case of Flow repository, if you get the unknown ID, then something is probably going wrong and there is nothing you can do, because Flow repo is the source of truth for Flows. Just log the error and `ack` in order not to get this message again?

SvenHoeffler:
@pnedelko Yeah, the second case should not happen in normal use, as the `flow.started`/`flow.stopped` events should never be spawned without having first been validated by the Flow Repository. I could potentially see a race condition happening, where a flow is deleted but there is still a `flow.started`/`flow.stopped` event floating around in the queue that is delivered some time after that deletion somehow. But yes, that seems like something that would have to be sorted out manually, and where a `nack` is of no use. I'll adjust it to call `nack` only for the case where the is DB being unreachable.

pnedelko:
@SvenHoeffler one more proposal/suggestion: would be nice to have some tests on what kind of information goes to the eventBus, as well as to test the event handlers.

SvenHoeffler:
@pnedelko Absolutely, that's still in progress. We're still investigating where/how to best intercept those calls.

services/flow-repository/app/api/controllers/startstop.js
philecs: @pnedelko & @SvenHoeffler I don't think we have one but we should define something in our API specs guidelines.

Any preferences? I would prefer `application/json`

SvenHoeffler: I think JSON is the option that the IAM is currently using as well (single object with a `msg` key), so we might as well standardise that.

philecs: I'll set up an issue to add this one to the guidelines.

PR #320, opened 2019-03-08
Update ServiceAPIGuidelines.md

philecs:


philecs:
```json
{
  "errors": [
    {},
    {}
  ]
}
```

philecs:
@SvenHoeffler @pnedelko lets discuss a format. What do you think about having an array of objects as format? So that we can show multiple error messages. We should also consider providing a structure for these objects.

SvenHoeffler:
@philecs I suppose that depends on whether there are components where an error does not outright stop the current operation. In case of the Flow Repo, generally any error stops the current request, as there's no chance of a successful further operation. So at least for that service there generally would not be instances where more than a single error would be spawned for a single normal operation.

Edit: Upon looking around some, it appears that this format is commonly used elsewhere as well. In one example where i could get an error to trigger, the overall result looked something like:
`{"errors":[{"error_message":"500 - Internal Server Error"}]}`

pnedelko:
[Here](https://nordicapis.com/best-practices-api-error-handling/) in 50% of examples there is `errors` array of objects. In [jsonapi](https://jsonapi.org/format/#document-top-level) is also used the `errors` array. Therefore I would propose:

```
{
  "errors": [
    {
      "message": "Required error message",
      "code": "Optional error code"
    }
  ]
}
```

Also it's worth to mention, that a proper http status code must be used by the APIs, so that an API doesn't respond with an error in it's payload and StatusCode=200.

philecs:
@pnedelko good proposal.

Agree that a proper status code is needed. Is anything missing from your pov in the current [guideline](https://github.com/openintegrationhub/openintegrationhub/blob/master/Guidelines/ServiceAPIGuidelines.md#http-requests) relating to status codes?

SvenHoeffler:
Agreed that the proposal by @pnedelko looks good.

@philecs As for status codes, I suppose we could mention in the guidelines for the sake of completeness that we use and abide by the official list of HTTP status codes (http://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml), and that the `errors` array should only be returned when the request returns with an error code (i.e. 4xx or 5xx)

philecs:
@pnedelko  and @SvenHoeffler I updated the file. Could you review pls?

PR #553, opened 2019-06-26
close before disconnect

fdegenhr:
fix for messages lost upon disconnect


philecs:
@pnedelko any objections against mergen the PR?

pnedelko (approved):
no objections

PR #737, opened 2019-09-05
Fixed typo in README.md: 8 GB of memory = 8192 MB instead of 8096 MB.

spipnl:
**What has changed?**

- Fixed typo in README.md: 8 GB of memory = 8192 MB instead of 8096 MB.

No careful review needed, as this is no code change.

---

[**Definition of Done**](https://github.com/openintegrationhub/openintegrationhub/blob/crudmonitoring/CONTRIBUTING.md#definition-of-done)


philecs:
@spipnl thanks for your contribution 👍

PR #986, opened 2020-08-17
Fix for #985

### LINKED ISSUE #985: Component Orchestrator Flows missing ELASTICIO_PUBLISH_MESSAGES_TO
It may be that the latest Component Orchestrator code (in Master) is not yet ready for production, but whenever I launch flows with the updated version, I get a failure inside the step node saying complaining about missing environment variables `ELASTICIO_CONTAINER_ID` or `ELASTICIO_PUBLISH_MESSAGES_TO`. I see that that former was commented out here: https://github.com/openintegrationhub/openintegrationhub/blob/master/services/component-orchestrator/src/drivers/kubernetes/KubernetesDriver.js#L309, but I don't see where `PUBLISH_MESSAGES_TO` is supposed to originate. I upgraded the node to the latest version of Sailor to see if that would fix it, but it did not.

Digging further, it appears that this commit (https://github.com/openintegrationhub/openintegrationhub/commit/b94462d89b5f2dcbea76b386f2b350a0dd51ad46) commented out that value from `createQueuesForFlowNode`, so it no longer got sent to the nodes.

https://github.com/openintegrationhub/openintegrationhub/blame/cf38bd64f19041c612c92a33cea391f2023300c2/lib/backend-commons-lib/src/QueueCreator.js#L313 

For the record, the reason I need the updated codebase is so I have the ability to increase resource allocation to different nodes, since we've had some get `OOMKilled`, as well as some other bugs that we've encountered that are not yet deployed to Docker hub.

mlb5000:
Fix https://github.com/openintegrationhub/openintegrationhub/issues/985

The big issue:
* `PUBLISH_MESSAGES_TO` was not set on `createQueuesForFlowNode`, breaking any Sailor nodes that tried to start up due to a requirement for that field.
* `DATA_ROUTING_KEY`, which is also required by Sailor, had been renamed to `OUTPUT_ROUTING_KEY`. I left both in place for now so the original author of that rename can determine the proper naming convention, or maybe leave these parallel paths in place for some time, with a deprecation path in mind assuming the renaming is desirable.

A couple smaller things:
* If `SNAPSHOTS_SERVICE_BASE_URL` is to be conditionally overridden, its default should be set first
* `CONTAINER_ID` and `WORKSPACE_ID`, although they are _technically_ irrelevant, are still required by some Sailor nodes


heggert:
We reworked parts of the container lib (sailor). See https://github.com/openintegrationhub/openintegrationhub/tree/master/lib/ferryman#installation

PR #989, opened 2020-08-17
Changes to allow for local development/deployment of Secret Service d…

mlb5000:
Changes to allow for local development/deployment of Secret Service docker containers

**What has changed?**

Previously, there was no real way to make changes to `lib/secret-service` and apply those changes to a Docker container that you could test/deploy on your own. This copies `Dockerfile` conventions from other services like `component-orchestrator`, but I'm pretty sure it's not 100%. 

This needs reviewed by someone familiar with the deployment processes just in case this Dockerfile is used by CI/CD. This was basically the only Dockerfile that only pulled its corresponding `lib` in by NPM instead of using the version alongside it in the repository.

* If you call `COPY lib/event-bus lib/event-bus` in the Dockerfile, it will break any usage of Event Bus from within `lib/secret-service` since the `Event` class is slightly different, thereby basically breaking certain APIs
* It looks like `service/src/conf/index.js` _should_ override the same configuration in `lib/secret-service/src/conf/index.js`, but I was wary to overwrite it as part of the Dockerfile process.


heggert:
Checkout https://github.com/openintegrationhub/openintegrationhub/pull/1019

Now you can start a local instance for development by running

`yarn start:nodemon`

in services/secret-service

PR #1054, opened 2020-09-17
format for real

blaggacao:
TL;DR : this is precursor code backed "generally accepted" canonicalization of k8s yaml files to empower file based diffing strategies to consolidate the different sources of truth `./minikube` / `./services`.

---

`kustomize cfg fmt` will probably the canonical way of formatting kubernetes resources in the future.

Additionally, my recommendation is to incorporate this command into pre-commit tooling (if desired). Over `prettier` it does ensure most appropriate semantic order of k8s resources.

From [its documentation](https://github.com/kubernetes-sigs/kustomize/blob/master/cmd/config/docs/commands/fmt.md):
> Field ordering roughly follows the ordering defined in the source Kubernetes resource definitions (i.e. go structures), falling back on lexicographical sorting for unrecognized fields.

This semantic ordering is of great help for consistency &mdash; I've seen a lot of naive copy+paste & _"oops, forgot this / that"_ during work on #1041 with `services/**/*.yaml`. This is a first step towards mitigation and resolution.

Please look at the individual commits to see with witch commands the results can be reproduced independently.

Note that:
- it catches some collateral `yml` files &mdash; I thought those are no big deal to just subsume under `kustomize cfg fmt`
- if `/minikube` or `/dev-tools` shall be spared, for example, please simply cherry pick the commits your interested in; I'd hope thought to get `./services/` and `./platform` clean and this technique already catches a few schema errors such as in #1053 and similar other previous commits

@weberjm As a relatively massive chore, it would be nice if this could be merged quickly during expected low turnover, e.g. this weekend?

---

After this canonicalization to get from here to `minikube`, one would have to apply the following patches:

```console
$ ls services | xargs -n1 -I{} sh -c 'diff -u services/{}/k8s/deployment.yaml minikube/4-Services/{}/k8s/deployment.yaml >> {}.patch'
$ diff -u services/iam/k8s/deployment.yaml minikube/2-IAM/deployment.yaml >> iam.patch

$ cat $(fd '\.patch') | wl-copy
```
<br/><br/>

<details>
<summary><b>Diff from here to <code>./minikube</code> state</b></summary>

```diff
--- services/app-directory/k8s/deployment.yaml	2020-09-17 20:28:33.169058824 -0500
+++ minikube/4-Services/app-directory/k8s/deployment.yaml	2020-09-17 20:28:33.165058779 -0500
@@ -1,10 +1,13 @@
-apiVersion: extensions/v1beta1
+apiVersion: apps/v1
 kind: Deployment
[...]

PR #1056, opened 2020-09-18
cleanup/fix: extensions/v1beta1 -> apps/v1

blaggacao:
&rarr; `rg -Nl extensions/v1beta1 services | xargs -n1 -I{} sed -i 's|extensions/v1beta1|apps/v1|g' {}`
&rarr; inspired by the diffs in #1054 

@weberjm A no-brainer which avoids API incompatibility surprises downstream. The current configuration does not use any feature exclusive to the beta version of the api &mdash; I'm not even sure if anything has been deprecated since the beta version.

### DIFF
--- a/services/app-directory/k8s/deployment/deployment.yaml
+++ b/services/app-directory/k8s/deployment/deployment.yaml
@@ -1,4 +1,4 @@
-apiVersion: extensions/v1beta1
+apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: app-directory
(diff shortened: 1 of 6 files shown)


SvenHoeffler (changes requested):
Good catch on these! We've already went and pulled up the apiVersion for a lot of services, but these must've slipped through the cracks.

Unfortunately, kubeval throws an error when I try to validate the new files. If I remember correctly, that's because apps/v1 requires the field `spec.selector.matchLabels.app` with the value of the app name, which yet isn't included in the affected files. See for example: https://github.com/openintegrationhub/openintegrationhub/blob/a7d6eaa50c3d6abec70cee9e139eb9dcb108c63f/services/flow-repository/k8s/deployment/deployment.yaml#L10

heggert:
Changes are too far behind the current master branch. Please provide additional fixes to ensure compatibility.

PR #1055, opened 2020-09-17
Outfactor inline envs into secret and configmap resources

blaggacao:
##### In reality, this targets the branch of PR #1054 instead of master, please take into account when reviewing.

Command history audit follows. Programs used:
- [`yq` - yaml cli maniuplator](https://mikefarah.gitbook.io/yq)
- [`jfq` - jsonata cli transformer](https://github.com/blgm/jfq)
- [`fd` - find alternative](https://github.com/sharkdp/fd)
- [`rg` - ripgrep (grep alternative)](https://github.com/BurntSushi/ripgrep)
- `kustomize` - for canonicalization

With those tools, the results should be reproducible, please note, that I executed every command individually and checked their outputs against git stage, then staged or discarded hunk by hunk or even line by line the desired results. This way, every step is clearly auditable. Please see also some fill ins I had to decide upon in commit IIIb/III below.

---

### `cd ./services`

- **commit I/III** &mdash; prepare

```console
$ fd -p 'deployment.yaml' | xargs -n1 -I{} sh -c 'mkdir $(dirname {})/deployment'
$ fd -p 'deployment.yaml' | xargs -n1 -I{} sh -c 'mkdir $(dirname {})/config'  # prepare outfactoring of configs and secrets
$ fd -p 'deployment.yaml' | xargs -n1 -I{} sh -c 'mv {} $(dirname {})/deployment/'
$ fd -p 'service.yaml' | xargs -n1 -I{} sh -c 'mv {} $(dirname {})/deployment/'
$ mv component-orchestrator/k8s/service-account.yaml component-orchestrator/k8s/deployment
```

---

- **commit II/III** &mdash; out-factor configmaps &mdash; users will be expected to regard them as pure _examples_.

```console
$ # capture env values which are note referenced already from a secret
$ ConfigMapJsonataExtractExpr='spec.template.spec.containers.env[value]'
$ # extrat values and push yaml array into target file
$ fd -p 'deployment.yaml' | xargs -n1 -I{} sh -c "source={}; target=\$(dirname \$(dirname {}))/config/configmap.yaml; yq r -j \$source | jfq -y "$ConfigMapJsonataExtractExpr" > \$target"
$ # remove empty file
$ rm ils/k8s/config/configmap.yaml
$ # prepend with configmap GVK
[...]

PR #1062, opened 2020-09-21
Outfactor secrets using envFrom (Open for discussion)

blaggacao:
This is a follow up to discuss possible changes in the way secrets are referenced. It carves the respective changes out of #1055.

The key benefits of this change:
- Secrets are declared in a [very straight forward way](https://github.com/openintegrationhub/openintegrationhub/blob/ae150a7607a393ca1be3181f2a194a33ee03cc4c/services/iam/k8s/config/secret.yaml) any developer will feel comfortable with
- Having secrets in a special file, they can be reasoned about more easily by automated tooling
- Comparing secret files side by side, it will be more easy to keep them up to date and spot inconsistencies.
- At some point automated tooling could auto-generate markdown docs from well documented `secret.yaml.example` files, keeping reference implementation and reference docs in sync.

As @weberjm pointed out, there are [several options](https://kubernetes.io/docs/concepts/configuration/secret/#using-secrets) to declare secrets. This PR adopts the `envFrom.secretRef` variant.

These secrets would be mere exemplifications, within the following folder structure:
```console
k8s
├── config
│   ├── configmap.yaml
│   └── secret.yaml.example
└── deployment
    ├── deployment.yaml
    └── service.yaml
```

Users would therefore create their own _Secrets_ and _ConfigMaps_ and would apply components with commands similar to this:
```console
$ kubectl apply -R services/<service-name>/k8s/deployment
```

If users also want to adopt configmaps:
```console
$ kubectl apply -R services/<service-name>/k8s/deployment
$ kubectl apply services/<service-name>/k8s/config/configmap.yaml
```

Since secrets are highly deployment specific the following way to deploy would _not_ make much sense:
```console
$ kubectl apply -R services/<service-name>/k8s/deployment
$ kubectl apply services/<service-name>/k8s/config/configmap.yaml
$ kubectl apply services/<service-name>/k8s/config/secret.yaml.example
```
Still viable is:
```console
$ kubectl apply -R services
```

Users can use any deployment tool of choice to automate and steer the deployment of those artifacts in any combination that suits their needs.


heggert:
Changes are too far behind the current master branch. Please provide additional fixes to ensure compatibility.

blaggacao:
@heggert I'm hoping this is not a request with serious intentions after almost 8 months past. :wink:

PR #1064, opened 2020-09-21
cleanup: Remove superflue sec context

blaggacao:
```yaml
        securityContext:
          privileged: false
```

is a [k8s default](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/#securitycontext-v1-core), so remove it to avoid confusion (as if it was set to false on purpose):

> Run container in privileged mode. Processes in privileged containers are essentially equivalent to root on the host. **Defaults to false.**

### DIFF
--- a/services/attachment-storage-service/k8s/deployment/deployment.yaml
+++ b/services/attachment-storage-service/k8s/deployment/deployment.yaml
@@ -50,8 +50,6 @@ spec:
         terminationMessagePath: /dev/termination-log
         terminationMessagePolicy: File
         imagePullPolicy: IfNotPresent
-        securityContext:
-          privileged: false
   strategy:
     type: RollingUpdate
     rollingUpdate:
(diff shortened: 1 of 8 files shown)


weberjm:
These PRs are being closed without merging. The requested changes do not have any bearing on the functionality of the system, and any perceived benefit from conciseness is minimal. Alternatively, there are several possible benefits by leaving these statements visible to other users. Not all users are expected to keep in mind all potential options for defining a K8s object. By leaving these visible, discoverability is increased for users. This is especially true in cases where the parameter is included because it has been previously set to a non-default value, and later reverted.

blaggacao:
@weberjm Perfectly reasonable. I updated https://github.com/openintegrationhub/openintegrationhub/issues/1073 accordingly. Thanks for this precedent decision on declaring vs not defaults.

PR #1065, opened 2020-09-21
cleanup: Remove superflue terminationmessage

blaggacao:
```yaml
        terminationMessagePath: /dev/termination-log
        terminationMessagePolicy: File
```

is a [k8s default](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/#container-v1-core), so remove it to avoid confusion (as if it was set on purpose):

-- | --
-- | --
terminationMessagePathstring | Optional:  Path at which the file to which the container's termination message  will be written is mounted into the container's filesystem. Message  written is intended to be brief final status, such as an assertion  failure message. Will be truncated by the node if greater than 4096  bytes. The total message length across all containers will be limited to  12kb. **Defaults to /dev/termination-log**. Cannot be updated.
terminationMessagePolicystring | Indicate  how the termination message should be populated. File will use the  contents of terminationMessagePath to populate the container status  message on both success and failure. FallbackToLogsOnError will use the  last chunk of container log output if the termination message file is  empty and the container exited with an error. The log output is limited  to 2048 bytes or 80 lines, whichever is smaller. **Defaults to File**.  Cannot be updated.

### DIFF
--- a/services/attachment-storage-service/k8s/deployment/deployment.yaml
+++ b/services/attachment-storage-service/k8s/deployment/deployment.yaml
@@ -47,8 +47,6 @@ spec:
           httpGet:
             port: 3002
             path: /healthcheck
-        terminationMessagePath: /dev/termination-log
-        terminationMessagePolicy: File
         imagePullPolicy: IfNotPresent
         securityContext:
           privileged: false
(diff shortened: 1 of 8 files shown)


weberjm:
Closing superfluous PR.

PR #1066, opened 2020-09-21
cleanup: Remove superflue terminationgraceperiod

blaggacao:
```yaml
      terminationGracePeriodSeconds: 30
```

is a [k8s default](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/#pod-v1-core), so remove it to avoid confusion (as if it was set to 30 on purpose):

-- | --
-- | --
terminationGracePeriodSecondsinteger | Optional  duration in seconds the pod needs to terminate gracefully. May be  decreased in delete request. Value must be non-negative integer. The  value zero indicates delete immediately. If this value is nil, the  default grace period will be used instead. The grace period is the  duration in seconds after the processes running in the pod are sent a  termination signal and the time when the processes are forcibly halted  with a kill signal. Set this value longer than the expected cleanup time  for your process. **Defaults to 30 seconds.**

### DIFF
--- a/services/attachment-storage-service/k8s/deployment/deployment.yaml
+++ b/services/attachment-storage-service/k8s/deployment/deployment.yaml
@@ -17,7 +17,6 @@ spec:
         app: attachment-storage-service
     spec:
       restartPolicy: Always
-      terminationGracePeriodSeconds: 30
       containers:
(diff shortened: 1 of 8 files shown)


weberjm:
Closing superfluous PR

PR #1067, opened 2020-09-21
cleanup: Remove superflue restartpolicy

blaggacao:
```yaml
      restartPolicy: Always
```

is a [k8s default](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/#pod-v1-core), so remove it to avoid confusion (as if it was set to _Always_ on purpose):

-- | --
-- | --
restartPolicystring | Restart  policy for all containers within the pod. One of Always, OnFailure,  Never. **Default to Always.** More info:  https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#restart-policy


weberjm:
Closing superfluous PR

Issue #1060, opened 2020-09-21
Provide central linting rules for yaml files


RobinBrinkmann:
We need to establish unified linting rules in order to prevent conflicts, decrease the number of flies/lines changed in commits and make it easier to review and work with the code. First we will only look at yaml e.g. by using https://github.com/adrienverge/yamllint.

---

**Acceptance Criteria:**
- [ ] Decide for appropriate tool/approach
- [ ] Align and set linting rules
- [ ] Publish and document

---

[**Definition of done**](https://github.com/openintegrationhub/openintegrationhub/blob/contributingGuidelines/CONTRIBUTING.md#definition-of-done)

blaggacao:
I'd suggest `kustomize cfg fmt` for k8s resource manifests. The big advantage is that it also canonicalizes the _order_ of k8s yaml fields. Currently there are several copies of similar manifests, at least until those are connected by transformers/code-generators to a single source of truth, it happens that changes are made in one site and forgotten in another. Being able to efficiently diff manifests (which requires canonical order) is a necessary workaround in the mean time.

For pure yaml, I regularly use `prettier` with a respective file filter. This is the most commonly used tool, it can also extend to format other files (but filters can restrain it from doing so). Excluding k8s manifests to avoid conflicting with more specialized tools would be as easy as this [file wide comment](https://prettier.io/docs/en/ignore.html#yaml).

Usually those would go into a pre-commit hook, which I usually configure only on `git push` so that local frequent committing is not discouraged ("speed programming").

blaggacao:
> Publish and document

Using git push hooks this part can be mostly spared sneaking in at established standard git workflows.

RobinBrinkmann:
@blaggacao thanks for the suggestion! That is an interesting approach we will be evaluating. Do you have a ```prettier cfg``` for yaml that you could share?

blaggacao:
#### Complete pre-commit yaml file with `prettier`, `shellcheck` & `shfmt`

```yaml
---
default_stages:
- manual  # can be run manually with `pre-commit run --hook-stage manual --all-files <hookname>`
- push # can be installed as pre-push hooks
exclude: "(^k8s/kustomize/)"
repos:
- hooks:
  - entry: "prettier --write --list-different"
    exclude: "^$"
    files: "\\.(css|less|scss|graphql|gql|html|js|jsx|json|md|markdown|mdown|mkdn|mdx|ts|tsx|vue|yaml|yml)$"
[...]

PR #1069, opened 2020-09-22
Add default yaml linter config (Open for discussion)

heggert:
Open for discussion


blaggacao:
I'd suggest in addition or as complement: `kustomize cfg fmt` &rarr; canonicalizes k8s resources with semantic knowledge about the resources (except for normalizing string quotes: `'` &rarr; `"`). I _think_ behind the scene it's sig-cli body which pulls the strings and decisions behind what _"canonical"_ means in this context.

heggert:
We have decided to postpone the further definition of the standards indefinitely.
Checkout current state: https://github.com/openintegrationhub/openintegrationhub/issues/1060

PR #1070, opened 2020-09-22
Add k7e handle

blaggacao:
**What has changed?**

- add innocuous `Kustomization` manifests (& very low maintenance! &mdash; they only ever change should the folder structure change in the future which might involve one of my suggestions anyway :wink: )
- the current api does not change
- additionally, those resources can be consumed downstream as Kustomization urlpaths.
- as a _"not officially supported"_ feature, no documentation is required. `Kustomization` gitops power users are supposed to know what they re looking for, the mere existence of those files is in itself more than enough documentation.

***this is not possible otherwise, without forking the entire repo &mdash; or similar terrible hacks which involve copying files around manually***

```
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
resources:
  - github.com/openintegrationhub/openintegrationhub/services/app-directory/k8s/deployment?ref=master
  - github.com/openintegrationhub/openintegrationhub/services/app-directory/k8s/config?ref=master
  - [...]
```

\* you'd never use `?ref=master`, of course.

No specifically careful review required. This does not change current behaviour. Naming those files `Kustomization` instead of the alternative `kustomization.yaml`, they entirely evict normal `kubectl apply -R` commands.

---

[**Definition of Done**](https://github.com/openintegrationhub/openintegrationhub/blob/crudmonitoring/CONTRIBUTING.md#definition-of-done)

### DIFF
--- a/platform/Kustomization
+++ b/platform/Kustomization
@@ -1,0 +1,11 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+  - ./namespaces.yaml
+  - ./grafana.yaml
+  - ./influxdb.yaml
+  - ./ingress.yaml
+  - ./mongodb.yaml
+  - ./rabbitmq.yaml
+  - ./redis.yaml
+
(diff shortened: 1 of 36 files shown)


RobinBrinkmann:
@blaggacao thanks again for introducing kustomize. We do see the overall value of it, but decided not to include it in the core open integration hub repos yet. We would have to maintain and document it as well as answering potential questions. We will rather focus our resources on building out new functionality. We also see kustomize moving closer to core k8s and would like to see how much of it becomes standard anyways.

PR #1092, opened 2020-10-14
Changed LOGGING_LEVEL to LOG_LEVEL

IoannisLafiotis:
**What has changed?**

- LOGGING_LEVEL to LOG_LEVEL
- INFLUXDB_LOGGING_LEVEL to INFLUXDB_LOG_LEVEL !!


platform/influxdb.yaml
```diff
@@ -21,7 +21,7 @@ spec:
         ports:
         - containerPort: 8086
         env:
-        - name: INFLUXDB_LOGGING_LEVEL
+        - name: INFLUXDB_LOG_LEVEL
```
weberjm: InfluxDB uses "logging" in its variable declarations

https://docs.influxdata.com/influxdb/v1.8/administration/config/#logging-settings

services/reports-analytics/test/ts/env.influxdb
```diff
@@ -1,5 +1,5 @@
 INFLUXDB_DATA_ENGINE=tsm1
-INFLUXDB_LOGGING_LEVEL=error
+INFLUXDB_LOG_LEVEL=error
```
weberjm: InfluxDB uses "logging" for variable declarations.

https://docs.influxdata.com/influxdb/v1.8/administration/config/#logging-settings

weberjm (changes requested):
While we are cleaning up our configuration to use a standard variable for logging, we also must be aware of external system requirements. 

The two commented files which refer to InfluxDB should be reverted to INFLUXDB_LOGGING_LEVEL

services/reports-analytics/test/ts/env.influxdb
IoannisLafiotis: Changed it back INFLUXDB_LOGGING_LEVEL according to the Docs of influx!

weberjm (approved):
Thx

PR #1114, opened 2020-10-26
Removed the logging-service from the mono-repo

IoannisLafiotis:
**What has changed?**

- Removed the logging service from the Monorepo
- Add the Logging service to its own repo in the OIH

There are no conflicting dependencies between the services but the logging service deployment may need some fixing after the pull request!


weberjm (commented):
👍

PR #1204, opened 2021-03-03
Gh release

weberjm:
Adding Github Release Action. Not yet tested (because it can only be tested live...)


weberjm:
Could look at the "act" application for running locally in the future https://github.com/nektos/act

PR #1205, opened 2021-03-03
Update Release Action Description

weberjm:
Only updated comments for developer assistance


weberjm:
Will handle in other branch....

PR #1239, opened 2021-04-14
Google Cloud Setup link was broken

devinbalkind:
added the correct link


RobinBrinkmann:
Thanks @devinbalkind for noticing and the quick fix!

PR #1259, opened 2021-06-30
Merge Secret Service Library into Service

weberjm:
As the Secret Service library defined a complete service, it is more beneficial to consolidate all functionality into the service folder. Should it become useful to implement a secrets library, this should be written for reusable functionality and inclusion across services.


heggert (approved):
Everything is fine. Thx for providing this refactoring!

PR #1281, opened 2021-08-05
added the mutable role object

IoannisLafiotis:
- Added the MutableRole schema


services/iam/doc/openapi.json
```diff
@@ -1105,6 +1105,23 @@
...
+          "name": {
+            "type": "string"
+          },
+          "permissions": {
+            "type": "array",
+            "items": {
+              "$ref": "#/components/schemas/Permissions"
+            }
+          },
+          "scope": {
```
weberjm: If you look at the model and the routes for the Roles, you will see which elements are actually accepted when updating or creating the role.

Example:
https://github.com/openintegrationhub/openintegrationhub/blob/b5dda021caa0e291169d34dce00266c1f74714dd/services/iam/src/routes/roles.js#L102-L107

As you can see in the snippet from the POST router here, the scope is not accepted in the request. I am not sure what scope is used for in roles, and this should probably become a separate issue. 'description' and 'isGlobal' however, are accepted.

weberjm (changes requested):

services/iam/doc/openapi.json
IoannisLafiotis: It exists in the membership as well! I am not sure where it is needed though!

services/iam/doc/openapi.json
```diff
@@ -1100,7 +1100,7 @@
               "$ref": "#/components/schemas/Permissions"
             }
           },
-          "scope": {
+          "description": {
```
weberjm, line 1118: Oh no! You touched the Role schema now 😄 

So, because the GET /roles and GET /roles/:id routes do not filter or alter the schema on return, the Role which is returned here can include any of the fields found in the role schema.

https://github.com/openintegrationhub/openintegrationhub/blob/419bf06cf6737e6685f32d1ba3866a2a65d64c6b/services/iam/src/models/schemas/role.js#L9-L36

And, because timestamps is true above, it will also return the fields "createdAt" and "updatedAt"

weberjm (changes requested):
I wasn't going to request any updates on other Schemas, because it was outside the scope of fixing the broken reference in the swagger doc. But now that you touched the Role schema, you should make it correct!

weberjm (approved):
This is ok for now. We haven't received any other external requests. If you compare with the actual response provided by the "/roles" endpoint, you will see not all fields are captured here. The other fields are optional, though, so I don't see a major problem at the moment. If you would like to add the other fields like _description_  or _isGlobal_, feel free. Otherwise, is ok to merge

PR #1337, opened 2021-10-22
Update ingress.yaml to use networking.k8s.io/v1

daveshah:
**What has changed?**
The `ingress.yaml` file has been modified in this PR to v1, moving off of the v1beta1 version of the file.

No.

N/A

**Release Notes**
Currently, when running `kubectl apply -f ./1-Platform` locally (as called out in the [Getting Started Guide](https://openintegrationhub.github.io/docs/3%20-%20GettingStarted/LocalInstallationGuide.html#basic-open-integration-hub-infrastructure-setup) `kubectl` fails with the following:
```
error: unable to recognize "1-Platform/ingress.yaml": no matches for kind "Ingress" in version "networking.k8s.io/v1beta1"
```
This change addresses this error by moving this file to the `v1` api version.

### DIFF
--- a/minikube/1-Platform/ingress.yaml
+++ b/minikube/1-Platform/ingress.yaml
@@ -1,4 +1,4 @@
-apiVersion: networking.k8s.io/v1beta1
+apiVersion: networking.k8s.io/v1
 kind: Ingress
 metadata:
   name: oih-dev-ingress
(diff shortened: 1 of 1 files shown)


weberjm:
Thanks for the contribution! As we mentioned on Slack, we have removed the minikube folder from the root of the repository, but you can find a similar, updated version under the dev-tools directory. This includes the updates to ingress which you have provided.

PR #1342, opened 2021-10-29
Updated Installation Link for Local Development

billgloff:
Existing link leads to a 404

**What has changed?**

- Change 1
- Change 2
- ...

_Please describe in short which change needs careful review, why and which files are affected_

- Change X
  - Why?
  - What files are affected?

**Release Notes**


RobinBrinkmann:
Thanks @billgloff for catching this and proposing an update.

PR #1345, opened 2021-11-11
remove hard-coded log level

### LINKED ISSUE #1344: Component/Connector LOG_LEVEL configuration too verbose
**Description**
The component LOG_LEVEL environment variable is hard coded to `trace` in the component orchestrator https://github.com/openintegrationhub/openintegrationhub/blob/master/services/component-orchestrator/src/drivers/kubernetes/KubernetesDriver.js#L380.  This is a too verbose setting for production environments.

**Solution Proposal**

Ideally this LOG_LEVEL is configurable or uses the component orchestrator LOG_LEVEL environment. At a minimum not setting the LOG_LEVEL environment variable will return it to the default `info` level set in ferryman.

I'm open to other solutions as well. If there is another setting available to do this, I may just be missing it in the code.

BirdHighway:
**What has changed?**

- The hard-coded component `LOG_LEVEL` environment variable has been removed from the component orchestrator
- `LOG_LEVEL` for components will now be determined by container environment variable, or default to Ferryman `info` if not specified
- Fixes issue #1344 

No

None

**Release Notes**
Component `LOG_LEVEL` no longer hard-coded by component-orchestrator


heggert (approved):

weberjm (changes requested):
Hi @BirdHighway, thanks for this contribution. In order to completely solve #1344, I think we should go ahead and make the log level configurable for pushing into the Global Component. 

Since I don't have access to the blendededge repos, I cannot make the changes myself. Here is what I would recommend:

In services/component-orchestrator/src/drivers/kubernetes/KubernetesDriver.js, add to the _prepareEnvVars():
```
envVars.LOG_LEVEL = this._config
            .get('COMPONENT_LOG_LEVEL')
            .replace(/\/$/, '');
```
after the envVar setup for DATAHUB_BASE_URL.

In services/component-orchestrator/config/default.json, add:
`"COMPONENT_LOG_LEVEL":"info",`

This should then clarify that the log level setting is to be passed to components, not for use in the orchestrator itself, and then be used in the running component. If you agree, just update the PR and I can merge today.

Also, would be helpful if you can rebase from the current base branch. 

Thanks!

weberjm (approved):

PR #1359, opened 2021-11-24
Minikube development fixes

winklerj:
**What has changed?**

- Added HOST_OIH_DIRECTORY to minikube start command for non-ARM based systems.
- Flow repository minikube: reduced the long delay for readiness and liveness probe

N/A

**Release Notes**

- Added HOST_OIH_DIRECTORY to minikube start command for non-ARM based systems.
- Flow repository minikube: reduced the long delay for readiness and liveness probe


weberjm:
Thanks for the PR. I wanted to ask about the addition of the HOST_OIH_DIRECTORY for non-ARM machines. This is used for the SourceCodeVolume. For non-ARM systems, the nfs connection to the host was supported, which is *supposed* to be faster: https://medium.com/skillshare-team/from-docker-compose-to-minikube-d94cbe97acda

When this was set up, NFS was not supported in the ARM systems, so the native mount was used. I'm not sure if later updates have either made NFS mounting available to ARM chips on Mac, or if NFS is no longer usable due to OS updates. Are you aware of this changing?

RobinBrinkmann:
@winklerj do you have any input regarding @weberjm comment? Would like to merge this, as it is open for quite a while, but we have to ensure the minikube setup then still works for all systems.

winklerj:
@RobinBrinkmann I believe @weberjm probably has better information on the HOST_OIH_DIRECTORY setting. I can remove that change from this PR and keep it to only adjusting the probe settings.

PR #1367, opened 2021-12-06
Add component rest api

heggert:
⚠️ PR moved --> https://github.com/openintegrationhub/openintegrationhub/pull/1381


jkopen (approved):

weberjm (approved):
The changes as made look technically correct. As we discussed personally, @heggert, our team's use cases would likely be somewhat different than yours, and would benefit from being able to call any function (especially triggers) through the REST API, and would then need an additional abstraction in the ferryman to handle the express routing parameters for each function.

Continuing down this line would also require other ferryman rewrites/updates/expansion, so that "value-add" functionality like snapshots can be done outside of the specific component function. This would allow the functions to be called via REST without affecting data sync for flows.

In the end, the process of standardization for a multi-tenant implementation would lead to decisions for the overall framework to be made, regarding which paths we support via configuration vs customization.

SvenHoeffler (approved):

PR #1375, opened 2022-01-10
Updated tenant path on open API doc

kerriannercrawford:
**What has changed?**
- Updated `paths` for `/api/v1/tenants/:id` to `/api/v1/tenants/{id}`. `/:id` was not being replaced by the real tenant id, rather the requests were being made directly to `/api/v1/tenants/:id`. 

No

N/A

**Release Notes**
Path for tenant API updated to replace `/:id` with the true tenant ID


SvenHoeffler (approved):
Good catch, thank you!

PR #1381, opened 2022-01-13
Add component rest api - 2nd

heggert:
As announced, I have made changes so that components can now also be accessed via REST API. Currently only a generic endpoint `http://node:port/process` lives here, which executes a predefined action.

Example action here
**dev-tools/test-component/lib/actions/exampleRESTAction.js**
```javascript
/* eslint consistent-return: "off" */
/* eslint max-len: 'off' */
/* eslint no-invalid-this: 0 no-console: 0 */

const log = require('../logger')

/**
 * This action will be called via REST API http://node:port/process and works like an express js middleware
 * Make sure that an appropriate response is sent. Unexpected errors are caught by an error handler
 *
 * @param {Function} req - express js middleware
 * @param {Function} res - express js middleware
 * @param {Function} next - express js middleware
 * @param {Object} actionParams - action parameters - { actionName: "string", secretId: "string", data: "object"}
 */

async function processAction(req, res, next, actionParams) {
  const { actionName, secretId, data } = actionParams
  log.debug('Going to process action', actionName)
  log.debug('Received data', secretId, data)
  /*
    Put logic here and send response
  */

  // examples
  if (data === null) {
    // use default error handler with 500 status code
    throw new Error('Oh noo!')
  } else if (!data) {
    // provide custom response error
    return next({
      status: 418,
      message: 'No data provided',
    })
  }

  if (data.custom) {
    return res.sendStatus(201)
  }

  // response data should be sent in following format
  /* 
    {
      data: {},
      meta: {} // optional
    }
  */
  res.send({
    data,
  })
}

module.exports.process = processAction

```

This action would be started via HTTP as follows
```
Request HTTP POST http://node:port/process
{
    "actionName": "exampleRESTAction",
    "secretId": "1234",
    "data": {
       "foo": "bar"
    }
}

Response HTTP 200 (Echo data)
{
    "data": {
       "foo": "bar"
    }
}
```
To make this possible, the following libs/services had to be extended
- Ferryman (core)
- Component Orchestrator (kubernetes driver)

To use this extension, the `descriptor` property of a component must be extended as follows:

```javascript
descriptor: {
    ...
    restAPI: true,
}
```
This causes the orchestrator to create a k8s **Service** in addition to the k8s **Deployment**, which allows access within the cluster via port `3001`. In addition, the Env `ENABLE_REST_API` is added to the component Envs. As soon as the component starts, the express server is also started and the endpoint mentioned above is available.


heggert:
See https://github.com/openintegrationhub/openintegrationhub/pull/1367#pullrequestreview-826396992

PR #1383, opened 2022-01-18
pass encryption key to /secrets patch handler

BirdHighway:
**What has changed?**

Encryption key is now passed to the secret service update secret route handler.

No

None

**Release Notes**

Secret service encrypts sensitive fields when a secret is updated via a PATCH request.


heggert (approved):
Thanks for providing a fix. The changes are optimal

heggert:
Now we just need to increase the version inside the package.json to create a new Docker image. I suggest that we declare this as a patch.

PR #1384, opened 2022-01-24
Smart assistance for flow recommendations

heggert:
With these changes we allow to identify already existing flows, which could then serve as an anonymized template for new flows.

Bildschirmfoto von 2022-01-24 14-30-17


heggert:
This branch will not be merged for the time being.

Issue #1174, opened 2021-01-20
Update Component Repository for Mapping and Action discoverability


weberjm:
To provide complete support for cross-domain mapping, and based on research done for the component generator (#1142), the Component Repository should be updated to provide enhanced discoverability of the available features of each Component. 

- To avoid duplication of work and failures to maintain synchronisation of component capabilities, the Component Repository should obtain information directly from the component. Action and Trigger information should come from the `component.json` file and schema information from the location decided in #1171.
- Component Repository should update its models to support schemas in its descriptor
- A decision and implementation must be made whether to obtain the component information on Create and Update of the component, or live from the component code on each call to the component repository.
---

**Acceptance Criteria:**
- [x] Component Repo API models all capabilities (actions, triggers, schemas)
- [x] Capability information is returned in API requests to Component Repo
  - [x] From its internal DB?
  - [x] From component code?
- [x] Capability information remains up-to-date with any component versioning or changes

---

weberjm:
One consideration for schema support: How should the component repository display schemas for APIs which allow for customization?
- Some APIs have a "free-form" field, such as _additionalFields_, which can be provided with an array or object for any user-customized parameters
- Other SaaS providers may have more complex support for more complex variations or custom objects, such as that [provided by Hubspot](https://developers.hubspot.com/docs/api/crm/crm-custom-objects)

Will this responsibility be put on the component creator / generator, or will the component repository have to be "intelligent" enough to process these objects?

PR #1385, opened 2022-02-02
Feat/add versioning components

russojrv-zz:
**What has changed?**

- We have implemented a new feature in the component-repository where we can have virtual components one level above the existing components and group them. Consequently, we can have multiple versions of one component.
- We added a new configuration collection where (for now) we can store one authClientId for tenant. In the future, we could add more configuration per component/tenant in this collection.

_It would be nice if you can take a look to the permissions and querys_

This may be related to this issue: [https://github.com/openintegrationhub/openintegrationhub/issues/1174](https://github.com/openintegrationhub/openintegrationhub/issues/1174)


lib/component-repository/package.json
```diff
@@ -28,11 +28,12 @@
     "eslint": "7.32.0",
     "eslint-plugin-mocha": "9.0.0",
     "eslint-plugin-node": "11.1.0",
+    "faker": "^5.5.3",
```
heggert: Please use this lib instead:
https://www.npmjs.com/package/@faker-js/faker

Why?
https://fossa.com/blog/npm-packages-colors-faker-corrupted/

lib/component-repository/src/routes/virtual-components/index.js
```diff
@@ -0,0 +1,178 @@
+const express = require('express');
+const asyncHandler = require('express-async-handler');
+const bodyParser = require('body-parser').json();
```
heggert: I would use the native functionality of express and do without the separate `bodyParser`. Basically, you only need to replace one line.
https://dev.to/taylorbeeston/you-probably-don-t-need-body-parser-in-your-express-apps-3nio

heggert (changes requested):

lib/component-repository/package.json
russojrv-zz: Replaced!

lib/component-repository/src/routes/virtual-components/index.js
russojrv-zz: Replaced to the native express functionality.

heggert (approved):

PR #1415, opened 2022-04-25
Added secret service preprocessor for OIDC auth client

winklerj:
**What has changed?**

- Secret service: Created an auth client preprocessor to map the id_token into the secret value
- Also, mapped the email in the JWT to the externalId like the Microsoft preprocessor

This functionality is modeled after other preprocessors that exist in the code and should be low risk

N/A

**Release Notes**
Added secret service auth client preprocessor for mapping an Open ID Connect (OIDC) id_token into the secret value created from the auth client.


sachmerz (approved):

RobinBrinkmann:
Thanks for your contribution @winklerj and review @sachmerz. 
PR approved and merged.

PR #1416, opened 2022-04-27
Secret Service: map id_token to externalId in oidc secret preprocessor

winklerj:
This is a follow up to #1415. I had not noticed that the id_token parameter is stripped from the secret before saving and was effectively the same as the `microsoft` preprocessor. I modified the preprocessor to set the externalId as the id_token.

**What has changed?**

- Modified id_token to be externalId in secret service `oidc` preprocessor

N/A

There should probably be an OpenID Connect specific auth type added as a follow up

**Release Notes**


winklerj:
If you would like to test I published a docker image winklerj/secret-service:22.1.1 with these changes.

sachmerz (approved):

PR #1419, opened 2022-05-11
Added orchestrator token context to log messages

### LINKED ISSUE #1418: Add back ComponentLogger context data
**Description**
Previously in version 1.x of the framework, RabbitMQ message headers would provide context information to the logs so that they could be correlated to an execution of the flow.  The logs could be queried by threadId to see the logs for all components in the flow. 

With the move to global components this information is now stored in the component orchestrator token but isn't added to log messages.  Additionally the threadId exists but isn't relevant because it will regenerate each time. The flowExecId needs to be used as a replacement for threadId

Currently this information is added to the ComponentLogger by ferryman here: https://github.com/blendededge/openintegrationhub/blob/master/lib/ferryman/lib/ferryman.js#L825

**Solution Proposal**

I'm proposing that the following additional context information below is added at line 825 above.  The values would be taken from the orchestrator token and I'm using this line for reference https://github.com/blendededge/openintegrationhub/blob/master/lib/ferryman/lib/ferryman.js#L638

- flowExecId
- flowId
- stepId
- userId
- tenant
- function

Please let me know if there is anything that needs added or removed from the list.

winklerj:
**What has changed?**
This PR fixes #1418. I have added the information listed in the issue

- Add component orchestrator token data to component log messages to provide context which aids in querying logs

N/A

- It doesn't look like tenant is populated in the orchestrator token and I don't see it in the logs
- There is a security vulnerability in the requestretry dependency. I didn't want to increment the version since it was several major versions out of date.

**Release Notes**
Added flowExecId, flowId, stepId, userId, tenant, and component function to component log messages for traceability.

### DIFF
--- a/lib/ferryman/lib/ferryman.js
+++ b/lib/ferryman/lib/ferryman.js
@@ -822,7 +822,10 @@ class Ferryman {
 
 
             const taskExec = new TaskExec({
-                loggerOptions: _.pick(incomingMessageHeaders, ['threadId', 'messageId', 'parentMessageId']),
+                loggerOptions: {
+                    ..._.pick(incomingMessageHeaders, ['threadId', 'messageId', 'parentMessageId']),
+                    ..._.pick(tokenData, ['flowExecId', 'flowId', 'function', 'stepId', 'tenant', 'userId'])
+                },
                 variables: stepData.variables,
                 services: {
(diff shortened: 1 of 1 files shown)


RobinBrinkmann:
Thanks @winklerj for the very useful addition!

PR #1424, opened 2022-06-14
Added option to include OAuth2 scope when refreshing token

winklerj:
The OAuth2 specifications says that the scope can optionally be sent in the request to refresh and access token https://datatracker.ietf.org/doc/html/rfc6749#section-6. Some services require that the scope be sent in order to successfully refresh the token. An option was added to the auth client configuration to specify whether the scope should be sent.
**What has changed?**

- Added a refreshWithScope boolean to auth client configurations. If true the scope is optionally added to the refresh_token request
- Replicated how predefined scope and scope from the secret are combined for consistency
- Fixed a missing dependency when running the secret service tests

Changes can be tested with winklerj/secret-service:22.1.1 on Docker Hub

N/A

N/A

**Release Notes**
- Added ability to send OAuth2 scope along with refresh token request


weberjm:
@winklerj If you update the branch to remove merge conflicts (I think it is due to an update to the package-lock), I can get this merged in.

winklerj:
@weberjm this PR has been updated to remove conflicts

Issue #1452, opened 2022-07-26
Webhooks HMAC Auth Not Working


BirdHighway:
**Describe the bug**
When using HMAC auth for a flow all requests sent to the webhooks services return a `500 Internal Server Error` error.

**Expected behavior**
When using HMAC auth for a flow:
- requests sent to webhook service should be rejected if they lack a valid HMAC header
- requests sent to webhook service should be accepted if they have a valid HMAC header

**To Reproduce**

- Create an API Key secret with a value of `12345`
- Create a flow that uses HMAC auth, specifying the id of the secret created above
- Start flow
- Send message to flow
  - Create message body `{"abc":123}`
  - Calculate HMAC for body using key created above, SHA-256 hash function, output to base64
    - [calculate hmac online](https://www.devglan.com/online-tools/hmac-sha256-online)
    - expected value `gHPCbdbiBH+2TbRx+jM9UC6snDdDxBrw9Bck4h7zewo=`
    - set as value under request header `X-Hmac-SHA256`
-  The request will be rejected

Secret Data (abbreviated)
```json
{
  "value": {
    "key": "12345"
  },
  "_id": "62dffbe8f1ed67f9eac3f9ca",
  "name": "HMAC Key",
  "type": "API_KEY"
}
```

Flow Data (abbreviated)
```json
{
  "name": "HMAC Example Flow",
  "graph": {
    "removed for brevity":"..."
  },
  "flowSettings": {
    "webhooks": {
      "hmacAuthSecret": "62dffbe8f1ed67f9eac3f9ca",
      "requireWebhookAuth": true,
      "hmacHeaderKey": "X-Hmac-SHA256"
    }
  }
}
```

Request:
```sh
curl --location --request POST 'http://webhooks.example.com/hook/62dffc6d667ffa180c1c86c0' \
  --header 'X-Hmac-SHA256: gHPCbdbiBH+2TbRx+jM9UC6snDdDxBrw9Bck4h7zewo=' \
  --header 'Content-Type: application/json' \
  --data-raw '{"abc":123}'
```

Response:
```http
HTTP/1.1 500 Internal Server Error
Date: Tue, 26 Jul 2022 15:51:54 GMT
Content-Type: application/json; charset=utf-8
Content-Length: 57
Connection: keep-alive
Vary: Origin
Access-Control-Allow-Credentials: true
ETag: W/"39-IvvnrLrqcQqmyk4cnGNn3OG58x4"

[...]

PR #1453, opened 2022-07-26
Webhooks HMAC Auth Fix

BirdHighway:
**What has changed?**

- HMAC data is sent to secret service for validation via `POST` request as a `base64` encoded string
- `DEFAULT_HMAC_ALGORITHM` is set to `sha256`
- HMAC is verified when specified by `flowSettings`, not based on presence / absence of HMAC header

No, other than making sure existing functionality is not affected by these changes.

No other known issues. Closes issue #1452 

**Release Notes**
Fixes issue with webhooks HMAC authentication


weberjm:
Robb and I had debated when to enforce the HMAC header and had decided on trying to auth with it whenever it is provided. I think what you have done is a better solution, that puts the enfocement power on the flow settings.

Regarding the switch on the rawBody, can you give a bit more info on what/why it was done? I thought we had both tested and things were working there. I do see that rawBody was previously set as a Buffer, and the process you are implementing here is
- Converting rawBody from a Buffer to a base64 string
- Passing to Secret Service, to validate HMAC
- parsing the rawBody back to a Buffer
- then converting the Buffer back to a String

The final step might also be unnecessary, since the `<hmac>.update()` function can already [accept Buffers](https://nodejs.org/api/crypto.html#hmacupdatedata-inputencoding). Did you run into specific instances where this is failing? i.e. is this a separate use case for a new service, where their body formatted fifferently, or was this a bug from the beginning?

BirdHighway:
Back in November Robb had asked me to verify that the webhooks auth was working, and I told him it was, but I don't think I tested it properly / I may have been mistaken about whether it worked as intended.

As far as the rawBody conversions - it sounds like I added in additional unnecessary work! Let me take a look to make sure I'm not missing something, but I think I probably just added changes in the process of debugging and didn't properly question whether they were necessary once I got it to work.

BirdHighway:
So `rawBody` is initially a buffer when we first have access to it in the webhooks service. We need to send this buffer to the secret service as part of a JSON request that includes other pieces of information that we need (like the `hmacValue` and `hmacAlgo`).

The output of calling `JSON.stringify()` on a buffer is an object that contains an array with the buffer data converted to an array of ints:
```json
{"type":"Buffer","data":[123,34,97,98,99,34,58,49,50,51,125]}
```

On the other end of the request, in the secret service, we parse the incoming JSON, select the `data` array and convert that to a buffer again with `Buffer.from(data)`.  This approach definitely works and has the advantage that it might be a little less convoluted. Here's a comparison of that approach with the pull request I made: https://github.com/blendededge/openintegrationhub/compare/webhooks-hmac...blendededge:openintegrationhub:webhooks-hmac-buffer

(The `<hmac>.update()` function can accept a buffer, but not a normal array, which generates the error: `The "data" argument must be of type string or an instance of Buffer, TypedArray, or DataView. Received an instance of Array.`)

The only real advantage I see to sending the `rawBody` as base64 encoded data is that it is more efficient. For example the data `{"abc": 123}` becomes:
- `"eyJhYmMiOiAxMjN9"` in base64 (16 bytes over the wire)
- `"123,34,97,98,99,34,58,32,49,50,51,125"` as an array (37 bytes over the wire)

If we weren't sending data over the network as a (JSON) string keeping everything as a buffer would be much more efficient. But with it having to be converted to a string I think it's less clear cut.

weberjm:
Thanks for the detailed review. I also don't see a huge advantage to choosing one over the other: a very slight processing time improvement from not having to encode and decode from base64, versus body size in the network transmission. It might just be a wash, there.

If you prefer to leave base64, then, I can approve it that way. If I am not mistaken, the removal of the toString() function inside the update() would still be beneficial, though, yes?

BirdHighway:
The current use of the `.toString()` method is necessary because the base64 has to be converted to utf-8 (the assumed original encoding of the data sent to the webhooks endpoint). Without converting `.toString()` the validation will fail. There may well be a smoother way to do this, but I'm not sure what it would be.

BirdHighway:
If the two approaches are roughly equal, I would vote for the approach that changes the existing code the least, which could be something like this:
https://github.com/openintegrationhub/openintegrationhub/compare/master...blendededge:openintegrationhub:webhooks-hmac-buffer

weberjm:
That's right. Even though both representations are the same string, they would give different HMAC values because of the encoding.

As I said above, I think the difference between processing a Buffer and a base64 string are probably minimal, so you are welcome to add in the last changes if that is your preference. Just let me know when the PR is ready, and I will merge it.

BirdHighway:
Okay thank you, it is ready to merge.

PR #1454, opened 2022-07-26
Secret service docs

BirdHighway:
**What has changed?**

Secret Service OpenAPI Docs changes:
- addition of auth client delete endpoint
- corrections to required properties of objects
- GET and PATCH requests for auth clients return `AuthClient`, not `MutableAuthClient`
- the version portion of the url is removed from the server url

These changes are required to enable working code to be generated via tools like [OpenAPI Generator](https://www.npmjs.com/package/@openapitools/openapi-generator-cli)

No

None

**Release Notes**
Minor fixes / improvements to Secret Service OpenAPI docs

### DIFF
--- a/services/secret-service/doc/openapi.json
+++ b/services/secret-service/doc/openapi.json
@@ -204,6 +204,31 @@
             "description" : "Invalid input"
           }
         }
+      },
+      "delete": {
+        "tags" : [ "Auth clients" ],
+        "summary" : "Delete a client",
+        "description" : "Delete a client",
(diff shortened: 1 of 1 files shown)


weberjm:
Merging doc fixes. This corrects incorrectly formed openAPI schemas. Note: it would have been less verbose to use a base auth client schema and then use an `allOf` key to merge the objects, rather than copying the fields into each discriminated AuthClient type. See https://swagger.io/docs/specification/data-models/oneof-anyof-allof-not/#allof

Issue #1451, opened 2022-07-25
Snapshot Service "getAll" function does not work without a flowExecId


weberjm:
If the Snapshots endpoint {snapshots-service}/snapshots/{flowId}/steps is called, and the flowExecId is not added as a query parameter, the endpoint will return no data, even if many snapshots exist for that flow. This is due to the way that flowExecId is added to the query for Mongo's `findAll()`, which then expect to find a result where the flowExecId is specifically `null`.  This should be changed to allow for multiple objects to be returned across flow executions.

weberjm:
As the "getAll" function was added specifically for flow execution scoped calls, there has not yet been a demonstrated need for a broader call to snapshots for a flow. Until such need exists, this issue is not necessary and will be closed.

Issue #1450, opened 2022-07-25
Snapshot deletion does not work for multi-emit components


weberjm:
In the library for Component Orchestrator, the decision to purge snapshots is made based on the `_checkFlowProgress()` function in [ComponentOrchestrator.js](https://github.com/openintegrationhub/openintegrationhub/blob/0378017fb45ed6e3d0d1c3e054e6bba92c266600/lib/component-orchestrator/src/ComponentOrchestrator.js#L318). In this function, the number of "started" components in a flow execution is checked against the number of "succeeded" components.

However, many components split returned arrays and emit multiple 'data' events which are received by the Orchestrator. Each time this happens, the "succeeded" value is increased, as well as the "started" value if there is a next step in the Flow. In the multi-emit components, the "succeeded" value never reaches the "started" value, because in the first step there is only one "start" count and multiple "succeeded". The snapshots will therefore never be purged.

Example: a component which returns 3 objects and executes steps in order
First component is running: started (1), succeeded (0)
First emit handled: started (2), succeeded (1)
Second emit handled: started (3), succeeded (2)
Third emit handled: started (4), succeeded (3)
Second component handles object 1: started (4), succeeded (4) <- DELETE snapshots
Second component handles object 2: started (0), succeeded (1)
Second component handles object 3: started (0), succeeded (2)

Additionally, because of the components processing their queues simultaneously, the snapshots may be purged at other times, or may never be purged due to the timing of Mongo transactions.

In a real-world flow example, a colleague had the following flow state:
started: 32, succeeded: 64

weberjm:
A similar problem may occur if a component fails to send either a "data" message (because of no results) or an "end" message

weberjm:
Fixes will provide an "end" message and only mark flows succeeded upon completion. This should allow for correct monitoring of flows and snapshots.

Issue #1449, opened 2022-07-25
Snapshot Lookups indeterminate if multiple snapshots exist


weberjm:
Since the addition of the `flowExecId` to the Snapshots service, in order to allow for "complex" flows including the Logic Gateway, it is possible for multiple snapshots to exist for a given `flowId` and `stepId`. As the query on Mongo is not sorted, Mongo returns values in [Natural Order](https://www.mongodb.com/docs/manual/reference/method/cursor.sort/#std-label-return-natural-order), which is an indeterministic method to fetch the appropriate or latest snapshot. 

To Mitigate:
- The snapshots service should return fetches to `getOne()` on the snapshots controller in descending order based on updated date.
- Based on other fixes for bugs created by `flowExecId`, other changes may be required, and will be updated here.

weberjm:
Snapshots are now sorted in lastUpdated order, and standard snapshots now save without flowExecId, which should reduce the overall number of snapshots stored, and eliminate incorrect snapshots loading.

PR #1461, opened 2022-09-08
FIxes for tracking Function completions through Flow Execution Lifecycle

weberjm:
Previous changes to the Snapshots service have broken one of the primary use cases of snapshots: tracking a "last updated" date or similar across many executions of a flow (in effect, a persistent snapshot). #1449 

Primarily:
- All snapshots were given executionIds 
- When an execution finished, these were deleted #1450 
- The tracking for execution component functions in Orchestrator relied on the 'data' event, which could be emitted 1..n times per component execution. This caused the "flow state" tracking to break. 

I have added the ability for the ferryman to also emit an 'end' event, for the Orchestrator to track flow states through completion of each execution. In addition, the Snapshots service needs to be able to handle both execution-scoped and flow-scoped snapshots. #1451


winklerj:
@weberjm Will this change require any changes to components? The previous changes to the queues and breaking change in ferryman caused the components ferryman version being in sync with the framework version deployed.  

We were able to make our components backwards compatibly by checking for the presence of the environment variable and determining which ferryman version to use. I'm just trying to gauge if we will need to do that again.

weberjm:
> @weberjm Will this change require any changes to components? 

@winklerj, I have been reviewing this today before I complete the merge (which will then be followed by a new release). Due to the nature of the change, the improvements do require both Ferryman (i.e. the Components) and the Orchestrator to be updated to correct the previously introduced problems, however there should be no adverse affects from being "out of sync."

- If Ferryman alone is updated, then new "end state" messages will be dropped by Rabbit and the current behavior will be maintained
- If Orchestrator alone is updated, then the previously indeterministic behavior of flow states will be changed slightly: 
  - Snapshots will *never* be purged, as opposed to possibly sometimes being purged
  - Flow States will not record "succeeded" components, as it looks only for `end` events, not `data` events now

Therefore, before I merge, I plan to mark Orchestrator and Ferryman as minor, rather than major, version updates. Happy to hear any other comments or concerns from you as well.

PR #1478, opened 2023-06-27
Feat/add user filters

jorgerusso:
**What has changed?**

- Added a filter in /users to be able to filter by id or username
- Added fromTemplate in the template model to know if this template has been copied from another one.

- Change Filter
  - Why? To have the possibility to filter the users array

- Change Template model
  - Why? To be able to know from which template has been copied

### DIFF
--- a/services/template-repository/app/models/schemas/flowTemplate.js
+++ b/services/template-repository/app/models/schemas/flowTemplate.js
@@ -77,6 +77,7 @@ const flowTemplate = new Schema({
   status: { type: String, default: 'draft' },
   cron: { type: String },
   flowSettings: {},
+  fromTemplate: { type: mongoose.Types.ObjectId },
 }, { collection: 'flowTemplates', timestamps: true });
 
 module.exports.flowTemplate = flowTemplate;
--- a/services/template-repository/app/models/schemas/flowTemplate.js
+++ b/services/template-repository/app/models/schemas/flowTemplate.js
@@ -77,7 +77,7 @@ const flowTemplate = new Schema({
   status: { type: String, default: 'draft' },
   cron: { type: String },
   flowSettings: {},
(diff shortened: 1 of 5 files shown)


services/template-repository/app/models/schemas/flowTemplate.js
```diff
@@ -77,6 +77,7 @@ const flowTemplate = new Schema({
   status: { type: String, default: 'draft' },
   cron: { type: String },
   flowSettings: {},
+  fromTemplate: { type: mongoose.Types.ObjectId },
```
weberjm: By adding "fromTemplate" field in the template, will we have problems passing the "fromTemplate" field on to the flow from the original template rather than the template that the flow is actually generated from? i.e. since the names are the same, is there an issue in the flow-repository of overwriting the id with this new id?

weberjm (commented):
Can you give me a little more detail on the requirements for this? I don't see a specific related issue, and I don't have in my head right now what the actual use cases for a user id and username filter are.

services/template-repository/app/models/schemas/flowTemplate.js
jorgerusso: Yes, sorry. We want to be able to copy templates from one user to another. I thought about this field to be able to follow from which one it was copied. But, yes, that’s a good point. What do you think about "copiedFrom"?

weberjm: That is also fine, and removes the chance of any copy errors (I wasn't sure it would happen, just wanted to make sure that you had checked 😄 ). Also happy with names like "sourceTemplate".

jorgerusso: Yep, I like it more as well!

pnedelko (approved):

PR #1480, opened 2023-07-04
Restart on channel close

pnedelko:
**What has changed?**

- Change 1
- Change 2
- ...

_Please describe in short which change needs careful review, why and which files are affected_

- Change X
  - Why?
  - What files are affected?

**Release Notes**


ryanlunka:
Hey guys! I just wondered if there was any ETA on this making it into a release? This would solve some significant challenges that come up with these RMQ disconnections, I think. Thanks a ton.

RobinBrinkmann:
good point @ryanlunka. we are wrapping this up and will release in the next couple days.

PR #1482, opened 2023-07-10
Secret service SESSION_AUTH type, add FORM_AUTH request type

winklerj:
**What has changed?**
- Add FORM_AUTH request type. Previously the body was only sent with the `application/json` content-type. Now the request fields can be submitted as the `application/x-www-form-urlencoded` content-type as well.
- Add expires property configuration. This is to enable the token to be reused if an expiration is present in the response.  This will reduce API call latency and potentially overwhelming the external API.
- Used MS Graph API with a client_credential grant type to test The FORM_AUTH will submit the request fields as application/x-www-form-urlencoded type. The docker image used to test is blendededge/secret-service:22.2.2

This slightly changes the secret service API by adding an optional expirationPath configuration. Also the schema for SESSION_AUTH needed to change to store the optional `expires` date time. Previously the refreshToken function returned only the token string which was stored in the secret. It now returns an object with the token and expiration date time specific to the SESSION_AUTH type

N/A

**Release Notes**

- Secret service: Updated the SESSION_AUTH type with the ability to submit session credentials as `application/x-www-form-urlencoded` form data instead of only `application/json`

### DIFF
--- a/services/secret-service/doc/openapi.json
+++ b/services/secret-service/doc/openapi.json
@@ -777,6 +777,9 @@
           "tokenPath": {
             "type": "string"
           },
+          "expirationPath": {
+            "type": "string"
+          },
           "endpoints": {
             "type": "object",
             "required": ["auth"],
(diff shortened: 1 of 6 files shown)


weberjm (approved):
Looks good to me.

Issue #1473, opened 2023-05-15
RabbitMQ messages not marked as durable


winklerj:
**Describe the bug**
When queues are created by OIH in RabbitMQ, they are correctly created as durable queues. However, unless the messages are marked as persistent, they are considered transient. This has the effect that messages will be lost if a RabbitMQ server in the cluster goes down or is restarted.

**Expected behavior**
Messages in RabbitMQ queues should be durable and are persisted across restarts of RabbitMQ servers

**To Reproduce**

1. Create a code component which uses the wait function and will allow messages to be held in queue.
2. Create a flow with the code component
3. Send several messages to the flow and allow messages to accumulate in the queue
4. Terminate the rabbitmq-server pod
5. Check that the messages are still in queue when the rabbitmq-server pod becomes available again

**Your OIH Setup Information**
Latest version of the OIH

**Additional context**
According the RabbitMQ documentation the queue and the message must be marked as durable: https://www.rabbitmq.com/tutorials/amqp-concepts.html#queue-durability. If you look at the settings for the amqplib package which is used https://amqp-node.github.io/amqplib/channel_api.html#channel_publish. You must specify `persistent` property for the message to be durable.

Here is where the message is published in the event-bus library: https://github.com/openintegrationhub/openintegrationhub/blob/master/lib/event-bus/src/RabbitMqTransport.js#L179
There may be other places. So, this is just an example.

winklerj:
I experimented with adding the persistent option to line 179 above. It seems like some messages in the `component-orchestrator` queue are not acknowledged once the messages are persistent.  More investigation needs to be done to find the root cause.

weberjm:
Hey @winklerj, I think this is complete now?

winklerj:
Yes, I think we can close it unless someone discovers a spot where messages are published that I missed

PR #1496, opened 2023-09-26
Make RabbitMQ messages durable persisted to disk

winklerj:
**What has changed?**

- Fix for #1473

N/A

I'm not sure if this covers all the places where OIH publishes messages to RabbitMQ.

**Release Notes**

Fix issue where RabbitMQ messages weren't persisted to disk


weberjm:
I await the GH Issues coming in where people expected the messages to be transient only 😄

PR #1499, opened 2023-10-04
HMAC auth on secret-service returns 413 https status code 

winklerj:
when body is larger than 100kb https://www.npmjs.com/package/body-parser#bodyparserjsonoptions

**What has changed?**

- I changed the default to 1mb and made the value configurable.
- I also noticed body-parser wasn't listed as a dependency in the package.json and added it

N/A

N/A

**Release Notes**
Made secret-service request body limit configurable

### DIFF
--- a/services/secret-service/package.json
+++ b/services/secret-service/package.json
@@ -21,6 +21,7 @@
     "@openintegrationhub/iam-utils": "*",
     "assert": "2.0.0",
     "base64url": "3.0.1",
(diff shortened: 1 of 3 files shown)


weberjm:
That is strange that body-parser wasn't listed. I guess because of the way the packages were built using workspaces, the library always existed anyway in `node_modules/`. I guess technically body-parser isn't needed anymore in express, but we'll go with it for the moment.

PR #1500, opened 2023-10-24
Allow to filter flows by user

pnedelko:
**What has changed?**

- Removed restriction for user filter in flows repo, so that users within the same tenant can get flows.

**Release Notes**

Allow ordinary users to filter flows by user.


weberjm (approved):
Seems simple enough. Have you tested to ensure no leakage between tenants? I believe the `buildQuery()` function in the mongo.js file of the same folder will filter that correctly, but I am not positive. (Do flows always get tenant added to the owners list?)

pnedelko:
Yes. `tenant` is always added to the flow owners. https://github.com/openintegrationhub/openintegrationhub/blob/db2443dce9c387d172778ecfbc064537fd609cd2/services/flow-repository/app/api/controllers/flow.js#L153

Issue #1407, opened 2022-04-01
Incorrect "meta" fields in response for several Services


weberjm:
**Bug**
The Flow-Repository, Template-Repository, and Audit-Log all use the Mongo function `estimatedDocumentCount()` to return the document `count` which is returned in the `meta` tag for GET requests, and helps with Pagination. https://github.com/openintegrationhub/openintegrationhub/search?q=estimateddocumentcount 

However, the behavior of `estimatedDocumentCount` is to use the Collection index to quickly estimate the total number of documents contained, regardless of any filters or queries (these are not supported as a parameter for the function.

**Expected behavior**
The API calls should always return the number of documents which are available to return to the user. This would mean using the `countDocuments()` function instead. 

**To Reproduce**

- Run a query on Flows without admin permissions. 
- Despite only seeing a limited number of Flows, the `meta.count` value will contain all flows for the system.

**Note**
Other services may potentially use the also-deprecated Mongoose `count()` function. This should also be replaced with `countDocuments()`

PR #1503, opened 2023-10-30
use countDocuments for counting flows

pnedelko:
**What has changed?**

- Use countDocuments method to calculate flows to get correct results


weberjm:
This is related to #1407 for this service.

